PR #1, opened 2021-06-08
Added bakery and baking examples from qua-libs

galwiner:


talshani:
let's add poetry to the mix
with black, pytest and flake8

PR #6, opened 2021-06-14
Added attributes for getting list of operations and inverse ops

arthurostrauss:
Moved the active reset to be just before RB sequence


examples/bakery_examples/RB_1_qubit/RB_1qubit.py
```diff
@@ -16,14 +15,16 @@
 RB_baked_sequences = RB.baked_sequences
 duration_trackers = RB.duration_trackers
 inverse_ops = RB.inverse_ops
+print(RB_sequences[0].operations_list)
```
galwiner, line 18: why do we need these print statements?

galwiner (approved):

PR #10, opened 2021-06-24
Corrected padding_method issue

arthurostrauss:
There was an issue occurring with the padding method when the number of 0s to be added was even, it is now corrected.


galwiner (approved):
We should start adding tests to this library. and then get CI here too

PR #17, opened 2021-08-08
Corrections related to wrong assertions and update of readme

arthurostrauss:
Fixed bugs pointed out when playing with amplitude modulation within the b.run(), added a test for it, and replaced error messages with assertion errors.

### DIFF
--- a/qualang_tools/bakery/bakery.py
+++ b/qualang_tools/bakery/bakery.py
@@ -339,7 +339,9 @@ class Baking:
         def remove_Op(q):
             if self.length_constraint is None:
                 if self._out:
-                    del self.config["elements"][q]["operations"][f"baked_Op_{self._ctr}"]
+                    del self.config["elements"][q]["operations"][
+                        f"baked_Op_{self._ctr}"
+                    ]
                     del self.config["pulses"][f"{q}_baked_pulse_{self._ctr}"]
                     if "mixInputs" in self.config["elements"][q]:
                         del self.config["waveforms"][f"{q}_baked_wf_I_{self._ctr}"]
@@ -347,10 +349,15 @@ class Baking:
                     elif "singleInput" in self.config["elements"][q]:
                         del self.config["waveforms"][f"{q}_baked_wf_{self._ctr}"]
                 else:
-                    raise KeyError("delete_baked_Op only available outside of the context manager "
-                                   "(config is updated at the exit)")
+                    raise KeyError(
+                        "delete_baked_Op only available outside of the context manager "
+                        "(config is updated at the exit)"
+                    )
             else:
-                raise Warning("Operation could not be deleted because baking object does not update the config")
+                raise Warning(
+                    "Operation could not be deleted because baking object does not update the config"
+                )
+
         if len(qe_set) != 0:
             for qe in qe_set:
                 remove_Op(qe)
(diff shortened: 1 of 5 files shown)


qualang_tools/bakery/bakery.py
```diff
@@ -9,20 +9,50 @@
...
+    config,
+    padding_method="right",
+    override=False,
+    update_config: bool = True,
+    baking_index: int = None,
+):
+    """
+    Opens a context manager to synthesize samples for arbitrary waveforms
+    :param config: config file
+    :param padding_method: Method to pad 0s to format the waveform to match hardware constraint
```
yomach, line 22: You need to write here what are the options for the padding_methods

qualang_tools/bakery/bakery.py
```diff
@@ -9,20 +9,50 @@
...
+    (>16 ns and multiple of 4)
+    :param override: Define if baked waveforms are overridable when using add_compiled feature
+    :param update_config: Define if baked waveform should be added within the input config file
```
yomach, line 25: What is the alternative?

qualang_tools/bakery/bakery.py
```diff
@@ -9,20 +9,50 @@
...
+    :param baking_index: index of a reference baking object to impose length constraint on new baked waveform
+    (useful for matching lengths when using waveform overriding in add_compiled feature)
+    :return: No return, config is updated if update_config is set to True,
```
yomach, line 28: If there is no return, don't add a :return: part...
You can write that part in the docstring itself above

qualang_tools/bakery/bakery.py
```diff
@@ -256,6 +319,24 @@ def _get_pulse_index(self, qe):
...
 
+    def get_waveforms_dict(self):
+        return self.override_waveforms_dict
+
+    def delete_baked_Op(self, qe: str):
+        """
+        Delete in the input config of the baking object the associated baked operation and
+        its associated pulse and waveform(s) for the specified quantum element
+        :param qe: quantum element
+        :return:
```
yomach, line 330: delete :return:

qualang_tools/bakery/bakery.py
yomach: Read about it in the readme. Add here that the default is true.

Is there any usecase in which this would be False but the baking_index won't be provided?

If the answer is no, then it might be easier to just get rid of this parameter...

yomach (changes requested):

qualang_tools/bakery/bakery.py
```diff
@@ -13,22 +13,20 @@ def baking(
     config,
     padding_method="right",
     override=False,
-    update_config: bool = True,
     baking_index: int = None,
 ):
     """
-    Opens a context manager to synthesize samples for arbitrary waveforms
+    Opens a context manager to synthesize samples for arbitrary waveforms. No return, config is updated if update_config is set to True,
```
yomach, line 19: There is no update_config now :)
You can write:
Opens a context manager to synthesize samples for arbitrary waveforms. The inputted config is updated, unless a a baking_index is given, in which case the generated waveforms can be retrieved using the get_waveform_dict() method, in a format readily plugable as an overrides argument

qualang_tools/bakery/bakery.py
```diff
@@ -37,11 +35,13 @@ def __init__(
         config,
         padding_method: str = "right",
         override: bool = False,
-        update_config: bool = True,
         baking_index: int = None,
     ):
         self._config = config
-        self.update_config = update_config
+        if baking_index:
```
yomach, line 41: The Baking index starts from 0 or from 1?
If it starts from 0, you have a bug.
If it starts from 1, it's cool

yomach (changes requested):

yomach (approved):

PR #22, opened 2021-09-05
Changed definition of baking alignment without any arguments

arthurostrauss:
Before:
b.align() would involve all quantum elements present in the configuration and upload a waveform of 0s for all of them
Now:
It only aligns elements that were previously called in within the baking context manager

### DIFF
--- a/qualang_tools/bakery/bakery.py
+++ b/qualang_tools/bakery/bakery.py
@@ -353,7 +353,14 @@ class Baking:
         return index
 
     def get_qe_set(self):
-        return self._qe_set
+        if self._out:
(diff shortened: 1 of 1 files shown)


yomach:
Is this properly documented?

arthurostrauss:
Actually the docstring attached to the method already made that statement but it was not true, it was padding 0s for all elements in the config regardless of their involvement within the baking.

yomach:
Then this is not a behavioral change, it is a bug fix. 
Next time make sure it is clear in the PR and in the commit messages

yomach (approved):

PR #29, opened 2021-12-06
improve CI actions and refactor package names

qguyk:


qualang_tools/bakery/xeb.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import List, Callable, Union
```
yomach, line 2: Why are these removed? They are not used?
If I remember correctly, @arthurostrauss added them because of an issue Caltech had...

arthurostrauss: Correct, those aliases worked and were used to assert the type of the expected arguments in all the methods. Caltech had the issue when the type was directly specified (list instead of List for example).

qguyk: that makes sense, but they are not used in this file. I've added linter so deleted the unused imports

yomach: I think that they were used in the type validation, not?

yomach: It is used in the bakery file, it is unneeded here.

yomach (approved):

PR #32, opened 2021-12-07
added config wrapper 

galwiner:
Added config wrapper and the config-gui

### DIFF
--- a/qualang_tools/config/__init__.py
+++ b/qualang_tools/config/__init__.py
@@ -2,5 +2,41 @@ from qualang_tools.config.integration_weights_tools import (
     convert_integration_weights,
     compress_integration_weights,
 )
+from qualang_tools.config.components import *
+from qualang_tools.config.primitive_components import *
+from qualang_tools.config.builder import *
 
-__all__ = ["convert_integration_weights", "compress_integration_weights"]
+__all__ = [
+    "convert_integration_weights",
+    "compress_integration_weights",
+    "Controller",
+    "ArbitraryWaveform",
+    "ConstantWaveform",
+    "DigitalWaveform",
+    "MeasurePulse",
+    "ControlPulse",
+    "Mixer",
+    "Element",
+    "MeasureElement",
+    "ConstantIntegrationWeights",
+    "ArbitraryIntegrationWeights",
+    "ElementCollection",
+    "ReadoutResonator",
+    "Transmon",
+    "FluxTunableTransmon",
+    "Coupler",
+    "Oscillator",
+    "Port",
+    "AnalogInputPort",
+    "AnalogOutputPort",
+    "DigitalInputPort",
+    "DigitalOutputPort",
+    "Waveform",
+    "Pulse",
+    "Operation",
+    "IntegrationWeights",
+    "Weights",
+    "DigitalSample",
+    "Matrix2x2",
+    "AnalogOutputFilter",
+]
--- a/qualang_tools/config/builder.py
+++ b/qualang_tools/config/builder.py
@@ -1,0 +1,236 @@
+from typing import Union, Dict
+
+from qualang_tools.config.configuration import *
+from qualang_tools.config.exceptions import ConfigurationError
+from qualang_tools.config.components import *
+from qualang_tools.config.parameters import isparametric
+
+
+class ConfigBuilder:
+    def __init__(self) -> None:
+        """A class used to build the configuration of an experiment with QUA"""
+        self.objects = []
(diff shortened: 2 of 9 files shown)


galwiner:
This PR is now just for the config wrapper

qualang_tools/config/__init__.py
```diff
@@ -4,3 +4,9 @@
 )
 
 __all__ = ["convert_integration_weights", "compress_integration_weights"]
+
```
qguyk, line 7: fix all imports to be not relative, put all imports in order (format)

qualang_tools/config/builder.py
```diff
@@ -0,0 +1,211 @@
+import inspect
+from abc import ABC
+from typing import Union, Dict, List
+
+from .configuration import *
+from .exceptions import ConfigurationError
```
qguyk, line 6: same here, better to use full path from source root

qualang_tools/config/components.py
```diff
@@ -0,0 +1,868 @@
+from typing import Union, List, Tuple
+
+from .primitive_components import *
+from .exceptions import ConfigurationError
+
+__all__ = ['Controller', 'ArbitraryWaveform', 'ConstantWaveform',
```
qguyk, line 6: __all__ should be defined in the init.py

qualang_tools/config/examples/resonator.py
```diff
@@ -0,0 +1,35 @@
...
+from config.configuration import *
+from config.components import *
+from config.builder import ConfigBuilder
+
+cont = Controller("con1")
+
+res = ReadoutResonator("res1", outputs=[cont.analog_output(0), cont.analog_output(1)],
+                       inputs=[cont.analog_input(0),cont.analog_input(1)],
+                       intermediate_frequency=2e6)
+res.lo_frequency = 4e9
```
qguyk, line 12: I wonder where those examples are supposed to be. 

Also, probably worth adding the examples as tests, to validate that the generated qua_config is correct.

qualang_tools/config/primitive_components.py
```diff
@@ -0,0 +1,158 @@
+from typing import Any, Dict, List, Tuple
+
+__all__ = ['Port', 'AnalogInputPort', 'AnalogOutputPort', 'DigitalInputPort',
```
qguyk, line 3: __all__ to init as well

qguyk (changes requested):

qualang_tools/config/examples/resonator.py
yomach: I think that the examples should be in the `qua-libs` git. 
I just finished moving the bakery examples there.
Please move it there (or convince me otherwise)

qualang_tools/config/builder.py
```diff
@@ -0,0 +1,211 @@
...
+            if isinstance(obj, Oscillator):
+                if obj.mixer == elm.name:
+                    objects.append(obj)
+            elif isinstance(obj, Element):
+                if obj.mixer.name == elm.name:
+                    objects.append(obj)
+        return set(objects)
+
+    @property
+    def qua_config(self):
```
yomach, line 204: Shouldn't the properties be together?
And isn't this confusing? people will use that instead of build?

qualang_tools/config/configuration.py
```diff
@@ -0,0 +1,130 @@
+import copy
+from typing import List, Union, Dict
+
+from .primitive_components import *
+from .components import *
+from .parameters import *
+
+class QMConfiguration:
```
yomach, line 8: We need docstrings

yomach (changes requested):
In general, docstrings needs to be added to various places, and maybe improved a bit in some.
I would also move the examples out of this folder.

I didn't check the usability at all, I guess this would be a work in progress.

qualang_tools/config/builder.py
SatyaBade12: users should only use `build` method, I removed `qua_config`

qualang_tools/config/examples/resonator.py
SatyaBade12: I added some basic tests (not exhaustive though) for transmon and resonator configs. The examples are moved to `config-builder-examples` branch in qua-libs

yomach (approved):

qguyk (approved):

galwiner:
@SatyaBade12 can you please Squash and merge this so you maintain the credit on this nice work?

PR #36, opened 2021-12-20
Config GUI

nikolaqm:
# Feature

This PR adds GUI for creating configuration. It allows:
* visualization of existing config
* edits of existing fields
* addition of components based on `ConfigBuilder` from `qualang_tools.config`.

# Use

`python -m qualang_tools.config.gui`

When edits are made, user can download two files `config_edits.py` that contains all the edits. If needed, user can manually change this `config_edits.py` file, e.g. to delete unwanted additions of components.

# Restrictions

* while configuratoin can be explored when being build from `components` of `ConfigBuilder`, currently it cannot be directly edited if the items are initially added as objects. On the other hand, if configuration is purely config variable (no component objects are used in building it), it can be edited through the GUI.

# Possible future extensions

* Object view (currently just config file is seen)
* Raw file view in GUI
* Support for parameters in GUI
* Back views (show who uses some resource in GUI)
* Allow adding new field in direct config editing using OpenAPI schema (currently schema is used just to show docs)
* Restrict more tightly field value edits in direct config editing (currently it's not checking beyond general data type for correctness of entered values)
* Parsing initial raw config as minimal object model (so that we can see in object view even existing configs build before ConfigBuilder is used)
* Implement lazy addition to allow deletion of objects from config builder (currently there is no option to delete added element other than manually editing corresponding config_edits.py file)


qualang_tools/config/gui.py
```diff
@@ -0,0 +1,124 @@
...
+        print(
+            "\tIf web browser does not open, please enter the address stated above manually."
+        )
+        print("\tStop GUI by pressing Ctrl+C")
+        serve(app.server, host=host, port=port_number)
+        # for development:
+        # app.run_server(host="127.0.0.1", debug=False, port=port_number)
+    except OSError as e:
+        timer.cancel()
+        if e.args[0] != 98:
```
qguyk, line 120: what is this 98?

qualang_tools/config/server/qua1_openapi.json
```diff
@@ -0,0 +1,550 @@
+{
```
qguyk, line 1: this is qua from qm-qua 0.3.4.
it might change. does any change in the json requires change here?

if you have to save it here, please rename it to reflect the version supported.

qualang_tools/config/server/upload.py
```diff
@@ -0,0 +1,165 @@
...
+
+configuration = setup.build(configuration)
+"""
+        )
+    return "Starting from empty configuration..."
+
+
+def init_edits_file():
+    with open(os.path.join(UPLOAD_DIRECTORY, "config_edits.py"), "w") as fp:
+        fp.write(
```
qguyk, line 156: can you add tests that validate files are created correctly and can be loaded to python?
(in the tests folder, maybe create a function that do it and call it from the callback.

qguyk (changes requested):
@nikolaqm 
looks good, I've added some comments
I've requested some specific tests, but if you can add more tests - it's better.

qguyk:
@galwiner 
please see how to run the GUI and make sure it fits the product requirements.
also validate names and location of the files

qualang_tools/config/gui.py
nikolaqm: That is error code for "address already in use" when server tries to open port on which there is already process listening. I catching just that specific case.

qguyk: is this error code part of an enum, or you can get the number from somewhere?
if not, please add the explanation as a comment

nikolaqm: Added comment b9c70e7a04bc29e246f5bdaba53bc0bc69df715f

qualang_tools/config/server/qua1_openapi.json
nikolaqm: The version of the schema that is saved here is already in the file ( "info" > "version" ). We don't allow selection currently of different schema, so we can just keep updated this file, without changing things on two places (file name and its reference in the code). Is this satisfactory for now (just one major qua version)?
Once there are multiple major versions we can have multiple files, but then it would make sense to support in GUI selection of appropriate.

qualang_tools/config/server/upload.py
nikolaqm: Added in 9ed79ba98943b27a46d7dfa55fbdff4f0470cd03

qualang_tools/config/server/qua1_openapi.json
nikolaqm: Now we download latest schema if possible (cfd000af8f1cf5041bf5596689e83a795a9ced64 3517089f8633691dbd19d78c298e0e5d14095b3b 42f689b43b7d26705c7b7444649dc159f4daaf49) , if not just use local copy

qguyk (approved):

PR #38, opened 2021-12-22
waveform tools - DRAG

KevinAVR:
Added DRAG waveforms to qualang_tools under config/waveform_tools


qguyk (changes requested):
@KevinAVR please add tests in the tests folder, that checks both functions output is correct

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+
+# Definition of pulses follow Chen et al. PRL, 116, 020501 (2016)
+
+
+def drag_gaussian_pulse_waveforms(
+    amplitude, length, sigma, alpha, detune, delta, substracted
+):
+    """
+    Function that returns DRAG pulses that compensate for leakage and also for the AC stark shift
+    that the leakage creates to the qubit frequency.
```
yomach, line 11: Add reference to the Thesis

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Envelope: gaussian
+
```
yomach, line 14: the amplitude of the gaussian, in volts

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+    :param amplitude: amplitude of voltage
+    :param length: pulse length
+    :param sigma: gaussian standard deviation
+    :param alpha: DRAG coefficient
+    :param detune: frequency shift to correct for AC start shift
+    :param delta: f_21 - f_10 - detune
```
yomach, line 20: what is delta? why do we have both alpha and delta if their only usage is alpha/delta?

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+    :param substracted: if you would like to use a substracted gaussian
+    """
+    t = np.arange(length, dtype=int)  # array of size pulse length in ns
+    gauss_wave = amplitude * np.exp(
+        -((t - length / 2) ** 2) / (2 * sigma ** 2)
+    )  # gaussian function
+    gauss_der_wave = (
+        amplitude
+        * (-2 * 1e9 * (t - length / 2) / (2 * sigma ** 2))
```
yomach, line 29: Are the units here correct? Do you get it in the correct scale?

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Envelope: gaussian
+
+    :param amplitude: amplitude of voltage
+    :param length: pulse length
+    :param sigma: gaussian standard deviation
+    :param alpha: DRAG coefficient
+    :param detune: frequency shift to correct for AC start shift
+    :param delta: f_21 - f_10 - detune
+    :param substracted: if you would like to use a substracted gaussian
```
yomach, line 21: :param bool substracted: If true, returns a substracted Gaussian to reduce high-frequency components due to the initial and final points offset. Default is true

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+# Definition of pulses follow Chen et al. PRL, 116, 020501 (2016)
+
+
+def drag_gaussian_pulse_waveforms(
```
yomach, line 6: add default values:
detune=0, substracted=True

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+    amplitude, length, sigma, alpha, detune, delta, substracted
+):
+    """
+    Function that returns DRAG pulses that compensate for leakage and also for the AC stark shift
+    that the leakage creates to the qubit frequency.
+
+    Envelope: gaussian
+
+    :param amplitude: amplitude of voltage
```
yomach, line 15: add types for all parameters:
:param double amplitude"
:param int length"
etc

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,63 @@
...
+    if substracted:  # if statement to determine usage of subtracted gaussian
+        gauss_wave = gauss_wave - gauss_wave[-1]  # subtracted gaussian
+    z = gauss_wave + 1j * gauss_der_wave * (alpha / delta)  # complex DRAG envelope
+    z *= np.exp(1j * 2 * np.pi * detune * t * 1e-9)  # complex DRAG detuned envelope
+    I_wf = z.real.tolist()  # get the real part of the I component of waveform
+    Q_wf = z.imag.tolist()  # get the imaginary part for the Q component of waveform
+    return I_wf, Q_wf
+
+
+def drag_cosine_pulse_waveforms(amplitude, length, alpha, detune, delta):
```
yomach, line 41: and everything for this one as well :)

yomach (changes requested):

yomach:
> @KevinAVR please add tests in the tests folder, that checks both functions output is correct

I'm not sure it's easy to add checks to this function...

qualang_tools/config/waveform_tools.py
KevinAVR: Yes, you do I have tested in my local files

qualang_tools/config/waveform_tools.py
KevinAVR: delta is the difference in energy between 2-1 and 1-0 energy levels. If you are using detuning it also will be included. The units of delta are in MHz, and alpha is unitless and usually range between 0 and 1 as reported in the literature. If you were to define just one value alpha/delta then you would have something in the range of 0 to 1e-6

qualang_tools/config/waveform_tools.py
KevinAVR: will do

qualang_tools/config/waveform_tools.py
KevinAVR: nice catch

KevinAVR (commented):
I have updated after your comments

qualang_tools/config/waveform_tools.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Envelope type: gaussian
+
+    :param float amplitude: amplitude in volts
+    :param int length: pulse length in ns
+    :param float sigma: gaussian standard deviation
+    :param float alpha: DRAG coefficient
+    :param float detune: frequency shift to correct for AC start shift
+    :param float delta: f_21 - f_10 - detune
+    :param bool substracted: if you would like to use a substracted gaussian
```
yomach, line 42: Change to:
:param bool substracted: If true, returns a substracted Gaussian, such that the first and last points will be at 0 volts. This reduces high-frequency components due to the initial and final points offset. Default is true.

yomach (changes requested):

qguyk:
> > @KevinAVR please add tests in the tests folder, that checks both functions output is correct
> 
> I'm not sure it's easy to add checks to this function...

@yomach does the thesis has an example of the result for specific parameters?

yomach:
> > > @KevinAVR please add tests in the tests folder, that checks both functions output is correct
> > 
> > 
> > I'm not sure it's easy to add checks to this function...
> 
> @yomach does the thesis has an example of the result for specific parameters?

Yes, given as an equation/figure...
Let me ask you this: does Scipy have a test checking their Gaussian Window function?

qguyk:
> > > > @KevinAVR please add tests in the tests folder, that checks both functions output is correct
> > > 
> > > 
> > > I'm not sure it's easy to add checks to this function...
> > 
> > 
> > @yomach does the thesis has an example of the result for specific parameters?
> 
> Yes, given as an equation/figure... Let me ask you this: does Scipy have a test checking their Gaussian Window function?

I think so: https://github.com/scipy/scipy/blob/f290778749003c09769bd21439620048fab2cdc5/scipy/signal/tests/test_windows.py#L340

yomach:
> > > > > @KevinAVR please add tests in the tests folder, that checks both functions output is correct
> > > > 
> > > > 
> > > > I'm not sure it's easy to add checks to this function...
> > > 
> > > 
> > > @yomach does the thesis has an example of the result for specific parameters?
> > 
> > 
> > Yes, given as an equation/figure... Let me ask you this: does Scipy have a test checking their Gaussian Window function?
> 
> I think so: https://github.com/scipy/scipy/blob/f290778749003c09769bd21439620048fab2cdc5/scipy/signal/tests/test_windows.py#L340
> 
> I would suggest taking the equation and getting results using Mathematica or MatLab so we can know that our function is correct

Ok, let's follow SciPy - @KevinAVR, can you please calculate this once for several values with several different parameters combinations (make sure you take both even and odd lengths), save this as lists, and then hardcode them into a test?

We can do this together if you want :)

qualang_tools/config/README_waveform_tools.md
```diff
@@ -0,0 +1,28 @@
+# Waveform Tools
+
+## Introduction
+---------------
+This package includes tools for creating waveforms. 
+
+## Functions
+------------
+* `drag_gaussian_pulse_waveforms` - Creates a gaussian based DRAG waveforms that compensate for the leakage and for the AC stark shift.
```
KevinAVR, line 9: there is an english typo here.

"Creates gaussian based DRAG waveforms...."

qualang_tools/config/README_waveform_tools.md
```diff
@@ -0,0 +1,28 @@
...
+* `drag_cosine_pulse_waveforms` - Creates a cosine based DRAG waveforms that compensate for the leakage and for the AC stark shift.
```
KevinAVR, line 10: Same english typo

"Creates cosine based DRAG waveforms..."

qualang_tools/config/__init__.py
```diff
@@ -3,10 +3,13 @@
     compress_integration_weights,
     plot_integration_weights,
 )
+from qualang_tools.config.waveform_tools import (
```
KevinAVR, line 6: what is the disadvantage to import *, vs only import the exact waveforms to use? is there any memory or delay issue? -- trying to learn here

qualang_tools/config/waveform_tools.py
```diff
@@ -1,90 +1,80 @@
...
-
 
 def drag_gaussian_pulse_waveforms(
-    amplitude, length, sigma, alpha, delta, detune=0, substracted=True
+    amplitude, length, sigma, alpha, delta, detuning=0, subtracted=True
 ):
     """
-    Function that returns DRAG pulses that compensate for leakage and also for the AC stark shift
-    that the leakage creates to the qubit frequency.
+    Creates a gaussian based DRAG waveforms that compensate for the leakage and for the AC stark shift.
```
KevinAVR, line 8: English typo

"Creates gaussian based DRAG waveforms..."

qualang_tools/config/waveform_tools.py
```diff
@@ -1,90 +1,80 @@
...
 
-    These DRAG waveforms has been implemented following the next Refs.:
+    These DRAG waveforms has been implemented following the next refs.:
```
KevinAVR, line 10: not a crucial comment, but during my Ph.D. we used "Refs." when referring to citations in the text // is personal choice

KevinAVR (commented):
I like your additions, I did not find typos in the physics

qualang_tools/config/__init__.py
yomach: It's considered better practice in general (If you import things you don't need, you waste memory).
I don't really agree with that when writing scripts, but for packages, I follow that.

qualang_tools/config/waveform_tools.py
yomach: Cool, changing this

yomach (approved):

yomach:
@qguyk Did everything we discussed, please take a look.
An unrelated test is failing, can you please try to help?

qguyk (approved):
@yomach @KevinAVR looks good
regarding the test - seems like something changed with the CI, couldn't find a quick fix and I don't want it to delay this PR any more, so marked the test as skipped but we should fix it

yomach:
@qguyk I don't like formating the tests because of what it does to lists. It makes it much less readable... We discussed it in the past, can we undone?

yomach:
Since there are contributions from both Kevin and me, I'll do a merge commit and not a squash.
Might squa

qguyk:
> @qguyk I don't like formating the tests because of what it does to lists. It makes it much less readable... We discussed it in the past, can we undo it?

@yomach take a look on it now

PR #49, opened 2022-01-09
chore - rename files

qguyk:


yomach:
I was wondering about it, we need to decide on the naming convention...
I was following the SDK: QuantumMachinesManger

qguyk:
> I was wondering about it, we need to decide on the naming convention... I was following the SDK: QuantumMachinesManger

I think it's better if we follow pep-8
https://www.python.org/dev/peps/pep-0008/#package-and-module-names

qguyk:
@yomach the test pass when i used a wrong table, can you please add a small test for the randomized_benchmark?

yomach (approved):
Please see my changes and approve :)

qguyk:
@yomach look good, lets squash and merge?

PR #74, opened 2022-03-04
VNA mode

TheoLaudatQM:
Introduction of the VNA mode enabling easy S-parameter and spectrum analyzer measurements with the OPX

### DIFF
--- a/qualang_tools/control_panel/README_vna.md
+++ b/qualang_tools/control_panel/README_vna.md
@@ -1,0 +1,105 @@
+# VNA mode
+
+## Introduction
+
+This module allows to configure the OPX as a VNA for a given quantum element (readout resonator for instance) and 
+operation (readout pulse for instance) already defined in the configuration. Once created, it has an API for 
+defining which measurements are to be run depending on the downconversion solution used (ED: envolope detector, 
+IR: image rejection mixer, IQ: IQ mixer).
+
+## Usage Example
+
+### Initializing the VNA and set up a S-parameter measurement
+This opens up a VNA using an existing configuration and initializes it for the quantum element _qe_ and operation _op_. 
+If the quantum element is a resonator, then its quality factor _Q_ can be specified for estimating its relaxation time.
+VNA measurements can then be added using the _add_ED()_ (envelope detector), _add_IR()_ (image rejection mixer) and 
+_add_IQ()_
+(IQ mixer) methods depending on the downconversion device used.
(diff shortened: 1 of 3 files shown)


yomach:
@TheoLaudatQM - Two things:
The build failed because of linting, run the command "flake8 qualang_tools" to see the error. In this case, I think you need a double backslash there to fix it, and maybe also to add an r before: `ylabel(r"$\\sqrt{...`
Change "quantum element" every to just "element" - elements don't have to be quantum (It's a change we're slowly making everywhere).
Change the "qe" variable to "element". and also change the "op" to "operation". We rather not use abbreviations in variable names.

yomach:
Looks good. There's a change that needs to be done to the readme files structure, I'll do it.

yomach:
The build also failed because of formating ("black")

TheoLaudatQM:
@yomach Okay, I reformatted it with black, sorry for that...

yomach:
No worries, we all forget to do it from time to time :)
I pushed some changes to your branch (Don't forget to pull!)
Please check my edits in the latest commit and approve them:
https://github.com/qua-platform/py-qua-tools/commit/4213433b64e308b9838067dd4c6ef1e870f7cdf5
I fixed a few typos, rearranged it such that there will be a separate readme for each tool, and renamed a few functions (This is mainly what you need to review).
In addition, I noticed that the description for the "dual" flag of "run_all" is different between the readme file and the function itself, can you please see which one is correct?
Once this is done, we'll merge it.

TheoLaudatQM:
To me everything seems fine, but I didn't see where I could approve...
I also updated the "dual" flag description in the last commit.

yomach:
Hi @TheoLaudatQM 
This is 95% done. Can you answer my last question and then we'll see if Guy Kerem has any other notes?

> Can there be more than two running at once?
> If so, I recommend we change it from "dual" to "parallel".

TheoLaudatQM:
Hi @yomach ,
Sorry I did not see your last question, now it can only perform two running at once in the spirit of doing S11 and S21 in parallel. I don't know if it would be useful to perform more than just two measurement in simultaneously here, but if it is needed I could add it easilly...

yomach:
@TheoLaudatQM Then we'll just keep it as is. Thanks.

yomach (approved):

qguyk (commented):
@yomach please manually test it with an opx, if all works i think we can merge it

yomach:
@TheoLaudatQM - Can you do a second check to make sure it works with the OPX after all of the small changes?

PR #83, opened 2022-03-20
Updated 2D tools, Rectangle data operation, data cleaning and Voronoi

asafr-quantum:
many updates to the interactive plot lib.
first, 2D graph support: copy-paste, delete etc. convert from 2d to 1d using "l"(lines), and back using "p" (for pcolor)
added "r" (rectangle) to select data regions and copy/ delete them. useful to clean data or copy only part of 1D/2D regions.
when using rectangle and single click - will select a single line from the 2D
Fit decay sine * exp now works great.
the output of fit now properly printed in the console for pycharm support
right-click on fit line will extend it across the visible axis.
Added "t" (transpose data) for 1D or 2D
added "v" (Voronoi). to split the screen using a voronoi graph, to count populations in qubit IQ blobs.
added commands (":"), for example, type ":log" anywhere to change the colormap to log scale in a 2D image
and many other bug fixes.

### DIFF
--- a/qualang_tools/addons/InteractivePlotLib.py
+++ b/qualang_tools/addons/InteractivePlotLib.py
@@ -14,7 +14,9 @@ import win32con
 from scipy.special import erf
 import pandas as pd
 import __main__ as _main_module
-
+import matplotlib.patches as patches
+from scipy.spatial import Voronoi
+import matplotlib.colors as Colors
 
 ipython = get_ipython()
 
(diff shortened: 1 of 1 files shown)


qguyk:
@asafr-quantum you are working on a stale branch. can you please create a branch on main and cherry pick the latest changes?

qguyk:
@asafr-quantum opened a new PR #84 instead

PR #94, opened 2022-04-15
Unit tests config builder

SatyaBade12:


yomach:
There are missing docstrings. Every function/class which is not private (starts with _) should have a docstring.
Other than this, this seems to be a very big change to all of the files... Will need to test the functionality...

SatyaBade12:
I added doc strings to the `Parameter` and `ConfigVars` classes concerned by this PR (let's create new issues for the remaining missing doc strings). Some of the changes done here are related to QUAM, but they also improve the overall code and I think they could be integrated irrespective of QUAM's current/future requirements.

yomach (approved):

qguyk:
@yomach we fixed some issues, and updated the changelog with the changes. It's breaking changes (need to be reflected by the version number). please approve the change again and then we can merge

qguyk (approved):

yomach (approved):

PR #96, opened 2022-05-15
initial state discriminator

yomach:
State discriminator for two states in the IQ plane


qualang_tools/analysis/discriminator.py
```diff
@@ -0,0 +1,133 @@
...
+        false_detections_var = np.sum(Ig > threshold) + np.sum(Ie < threshold)
+    else:
+        false_detections_var = np.sum(Ig < threshold) + np.sum(Ie > threshold)
+    return false_detections_var
+
+
+def two_state_discriminator(Ig, Qg, Ie, Qe, b_print=True, b_plot=True):
+    """
+    Given two blobs in the IQ plane representing two states, finds the optimal threshold to discriminate between them
+    abd calculates the fidelity. Also returns the angle in which the data needs to be rotated in order to have all the
```
TheoLaudatQM, line 17: Typo: and

TheoLaudatQM (approved):
I just quickly tested the rotations of the IQ blobs and it seems to behave as expected so to me it is fine !
Do you have ideas about what other functions will belong to this library? (gef discriminator  guess, ...)

yomach:
@TheoLaudatQM I added the tests and finished readme, please review again and see if you have any comments

qualang_tools/analysis/discriminator.py
```diff
@@ -0,0 +1,131 @@
...
+        Unexpected output will be returned in other cases.
+
+
+    :param float Ig: A vector containing the `I` quadrature of data points in the ground state
+    :param float Qg: A vector containing the `Q` quadrature of data points in the ground state
+    :param float Ie: A vector containing the `I` quadrature of data points in the excited state
+    :param float Qe: A vector containing the `Q` quadrature of data points in the excited state
+    :param bool b_print: When true (default), prints the results to the console.
+    :param bool b_plot: When true (default), plots the results in a new figure.
+    :returns: A tuple of (angle, threshold, b_excited_is_larger, fidelity, gg, ge, eg, ee).
```
TheoLaudatQM, line 31: b_excited_is_larger is not returned anymore

yomach: nice catch :)

TheoLaudatQM:
Looking at the test I am wondering if it would not be useful to add a function that rotates the IQ blobs according to the angle calculated with two_state_discriminator, like Ig_rotated, Qg_rotated, Ie_rotated, Qe_rotated = rotate_iq_blobs(Ig, Qg, Ie, Qe), when we will implement it in other scripts, what do you think?

yomach:
Do you see any use for that?
The idea is to put that angle into the configuration file, and then use the rotated integration weights. This means that the data would already come out rotated...

TheoLaudatQM:
That is true, I do not know if there is a use-case for which someone would want to run this calibration in a more complex sequence where they would need to rotate the blobs on the fly... If you don't think it is relevant then you can forget it :)

yomach:
@qguyk Any comments before I merge?

PR #97, opened 2022-05-16
utils functions

TheoLaudatQM:
utlis functions to make qua programs easier and shorter


qualang_tools/utils/README.md
```diff
@@ -0,0 +1,53 @@
+# QUA Utils
+
+QUA utils is a collection of small Python or QUA functions and classes which make QUA programs clearer and easier.
+
+It is meant to be updated regularly so feel free to share any comments and suggestions following these
+ [instructions](https://github.com/qua-platform/py-qua-tools#support-and-contribution).
```
yomach, line 6: Better to link directly here:
https://github.com/qua-platform/py-qua-tools/blob/main/CONTRIBUTING.md

qualang_tools/utils/README.md
```diff
@@ -0,0 +1,53 @@
+# QUA Utils
```
yomach, line 1: Need to think on the name :)
I actually think it's fine, but do we want to have everything here? what decides what goes here and what else where?

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        self.data_list = data_list
+        self.mode = mode
+        self.results = []
+        self.data_handles = []
+        self.res_handles = job.result_handles
+        if mode == "live":
+            for data in self.data_list:
+                if hasattr(self.res_handles, data):
+                    self.data_handles.append(self.res_handles.get(data))
+                    self.res_handles.get(data).wait_for_values(1)
```
yomach, line 46: you just got it a a line above :)

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,78 @@
...
+                else:
+                    raise Warning(f"{data} is not saved in the stream processing.")
+
+    def format_(self, data):
```
yomach, line 50: it's an internal function, not to be used by customers, so prefix it with _:
`def _format(...)`

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        if type(data) == np.ndarray:
+            if type(data[0]) == np.void:
+                if len(data.dtype) == 1:
+                    data = data["value"]
```
yomach, line 54: Are you handling all cases?
1. just .save()
2. just .save_all()
3. timestamps().save()
4. timestamps().save_all()
5. with_timestamps().save()
6. with_timestamps().save_all()

yomach (changes requested):

qualang_tools/utils/README.md
TheoLaudatQM: Yes we really need to discuss this point soon

qualang_tools/utils/qua_utils.py
TheoLaudatQM: I am sure you're right, but I don't get it :)

qualang_tools/utils/README.md
TheoLaudatQM: Done

qualang_tools/utils/qua_utils.py
TheoLaudatQM: Oh right I inverted it..., done !

qualang_tools/utils/qua_utils.py
yomach: It's not that critical...
You can change the 2nd line to be 
`self.data_handles[-1].wait_for_values(1)`
This would reduce the overhead of doing `self.res_handles.get(data)` twice

yomach: or 
```
data_handle = self.res_handles.get(data)
self.data_handles.append(data_handle)
data_handle.wait_for_values(1)
```

TheoLaudatQM: I see, good point indeed !

qualang_tools/utils/qua_utils.py
yomach: great!

qualang_tools/utils/qua_utils.py
```diff
@@ -1,6 +1,77 @@
 import numpy as np
```
yomach, line 1: This is without any logic. The idea is to make sure that the result is the same as numpy `arange`, which currently does not happen...
Cases where this would break:
a_min = 0.1, a_max = 1.0, da = 0.2 - Because +da/2 would give us 1.1, and it **might** be included...
a_min = 10, a_max = 20, da = 1 - Because it is an integer.

There are many more examples, and some are listed on the confluence page:
https://quantum-machines.atlassian.net/wiki/spaces/PLAN/pages/1144389988/Improve+for+behavior

This is a very big effort, I suggest to move it to a separate PR

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    assert len(array) > 0, "The array must be of length >0"
+
+    if np.isclose(np.std(np.diff(array)), 0):
+        increment = "lin"
+    elif np.isclose(np.std(array[1:] / array[:-1]), 0):
+        increment = "log"
+    else:
+        raise ValueError(
+            "The spacing of the input array must be linear or logarithmic."
```
yomach, line 20: I would add: Please use with `for_each_()` for scanning arbitrary values

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,237 @@
...
+def qua_arange(var, start, stop, step):
+    """
+     Function parametrizing the QUA `for_` loop from the numpy.arange() syntax.
+    :param var: the QUA variable that will be looped over (int or fixed).
+    :param start: Start of interval. The interval includes this value. Must be a Python variable.
+    :param stop: End of interval. The interval does not include this value. Must be a Python variable.
+    :param step: Spacing between values. Must be a Python variable.
+    :return: QUA for_ loop parameters (var, init, cond, update) as defined in https://qm-docs.qualang.io/api_references/qua/dsl_main?highlight=for_#qm.qua._dsl.for_.
+    """
+
```
yomach, line 86: need to add a validation that:
var is qua variable
start, stop and step are not

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,237 @@
...
+    :param var: the QUA variable that will be looped over (int or fixed).
+    :param start: The starting value of the sequence. The interval includes this value. Must be a Python variable.
+    :param stop: The end value of the sequence. The interval includes this value within the 2**-28 fixed point accuracy. Must be a Python variable.
+    :param num: Number of samples to generate. Must be a Python variable.
+    :return: QUA for_ loop parameters (var, init, cond, update) as defined in https://qm-docs.qualang.io/api_references/qua/dsl_main?highlight=for_#qm.qua._dsl.for_.
+    """
+    assert (
+        np.abs(start) <= 8 and np.abs(stop) < 8
+    ), "fixed numbers are bounded to [8, 8)"
+    assert type(num) is int, "The number of samples must be a python integer"
```
yomach, line 121: this would fail for a numpy int

yomach (commented):
I might separate this into two different branches and PRs.
One for the `for_` methods, and a second for the fetching methods.
Maybe also separate into two files?

qualang_tools/utils/qua_utils.py
TheoLaudatQM: Good point, done!

qualang_tools/utils/qua_utils.py
TheoLaudatQM: Solved

qualang_tools/utils/qua_utils.py
TheoLaudatQM: Done!

TheoLaudatQM:
> I might separate this into two different branches and PRs. One for the `for_` methods, and a second for the fetching methods. Maybe also separate into two files?

I understand why two different PRs, but why two different files?

qualang_tools/utils/qua_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        if self.mode == "wait_for_all":
+            self.res_handles.wait_for_all_values()
+            for data in self.data_list:
+                if hasattr(self.res_handles, data):
+                    self.results.append(
+                        self.format_(self.res_handles.get(data).fetch_all())
+                    )
+
+        elif self.mode == "live":
```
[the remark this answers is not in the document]
TheoLaudatQM: I think that we can close this PR now if it is okay and I will add it later after some tests (it is on my to-do list :) ), what do you think?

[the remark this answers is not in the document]
yomach: Yes

yomach:
> > I might separate this into two different branches and PRs. One for the `for_` methods, and a second for the fetching methods. Maybe also separate into two files?
> 
> I understand why two different PRs, but why two different files?

I think `qua_util` is a general name. We can decide to put everything there, but does it makes sense?
If someone tries to import something, it needs to make sense.

`from qualang_tools.util.qua_utils import result_tools`
`from qualang_tools.util.qua_utils import interrupt_on_close`
`from qualang_tools.util.qua_utils import qua_arange, qua_list, qua_linspace`

So a few points:

1. `qualang_tools.util.qua_utils` is a bit of double. Actually triple. You're referencing QUA twice and util twice
2. Instead of `util`, maybe `utility`? making an abbreviation only makes sense when you know about it
3. Instead of `qua_util` we need something else. Maybe "results_tools" for the result tools and "for_tools"? or "list_tools"?
4. Maybe the result tools can be in the base (directly in `qua_util`)?

All of these points are open for discussion

TheoLaudatQM:
> > > I might separate this into two different branches and PRs. One for the `for_` methods, and a second for the fetching methods. Maybe also separate into two files?
> > 
> > 
> > I understand why two different PRs, but why two different files?
> 
> I think `qua_util` is a general name. We can decide to put everything there, but does it makes sense? If someone tries to import something, it needs to make sense.
> 
> `from qualang_tools.util.qua_utils import result_tools` `from qualang_tools.util.qua_utils import interrupt_on_close` `from qualang_tools.util.qua_utils import qua_arange, qua_list, qua_linspace`
> 
> So a few points:
> 
> 1. `qualang_tools.util.qua_utils` is a bit of double. Actually triple. You're referencing QUA twice and util twice
> 2. Instead of `util`, maybe `utility`? making an abbreviation only makes sense when you know about it
> 3. Instead of `qua_util` we need something else. Maybe "results_tools" for the result tools and "for_tools"? or "list_tools"?
> 4. Maybe the result tools can be in the base (directly in `qua_util`)?
> 
> All of these points are open for discussion

I totally get your point, here are my comments:

- I agree that `from qualang_tools.utils.qua_utils import ...` is cumbersome and repetitive. 
- I think that `utils` is very generic and used in every programming language so people know what it is
- In this context, I agree that we should separate the different functions into different files. Especially if we want to put more and more things there in the future. adding tools seems also a bit redundant since we already import from qualang_tools so I was thinking about something like (`from qualang_tools.utils.loops import qua_arange` and `from qualang_tools.utils.results import fetching_tool` (in case we want to add a class acting on the results specifically) and `from qualang_tools.utils.plots import interrupt_on_close`)

yomach:
I'm good with your suggestions.
Again, I would pass them through Guy Kerem, he usually have good insights as well :)

yomach:
BTW, regarding QPlot:
It's something which only works in very specific IDEs, so the idea was to put it into a folder named "experimental", which is common for "not fully baked" features in packages.
However, "experimental" has a totally different meaning for us, so after a long discussion it settled into "addons"...  (Maybe not the best name).
Because of the huge number of dependencies, it requires, and the fact that simply importing them might not work in a Linux system, when you do pip install qm-qua it doesn't automatically install them...
If we want to get QPlot out of the "addons", we need to do a lot of testing, and also see if we can lighten on the requirements.

qualang_tools/loops/loops.py
```diff
@@ -0,0 +1,205 @@
...
+    """Function parametrizing the QUA `for_` loop from the numpy.logspace() syntax.
+
+    :param var: the QUA variable that will be looped over (fixed).
+    :param start: The starting value of the sequence. The interval includes this value. Must be a Python variable.
+    :param stop: The end value of the sequence. The interval includes this value within the 2**-28 fixed point accuracy. Must be a Python variable.
+    :param num: Number of samples to generate. Must be a Python variable.
+    :return: QUA for_ loop parameters (var, init, cond, update) as defined in https://qm-docs.qualang.io/api_references/qua/dsl_main?highlight=for_#qm.qua._dsl.for_.
+    """
+    # Check QUA vs python variables
+    if not hasattr(var, "_exp"):
```
yomach, line 184: I don't this `hasattr` is the correct way to go. It's too general.
Do: `isinstance(var, _Expression)`

qualang_tools/loops/loops.py
```diff
@@ -0,0 +1,205 @@
...
+        raise Exception("The first argument must be a QUA variable.")
+    if hasattr(start, "_exp") or hasattr(stop, "_exp") or hasattr(num, "_exp"):
```
yomach, line 186: To check if something is a number, use:

`if isinstance(start, (np.generic, int, float))`

This check against numpy (number, it will not catch an array), an int or a float

qualang_tools/loops/loops.py
```diff
@@ -0,0 +1,205 @@
...
+        raise Exception("The for loop arguments must be python variables.")
+    # Check for fixed number overflows
+    if not (-8 <= start < 8) and not (-8 <= stop < 8):
+        raise Exception("fixed numbers are bounded to [-8, 8).")
+    # Check if num is integer
+    if not float(num).is_integer():
```
yomach, line 192: I wondered if his would work if someone put a QUA variable, and it would fail.
Trying to do "float" on a QUA variable returns an error, so instead of the nice error you wrote, they would get a generic one...
All of these need to change into `isinstance` as I wrote above

qualang_tools/results/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+fig = plt.figure(figsize=(15, 15))
+
+while job.result_handles.is_processing():
+    # Live plotting
+    I, Q, Ie, Qe, Ig, Qg = my_results.fetch_all()
+    ...
+```
+
+## progress_counter
```
yomach, line 32: nice!!

qualang_tools/results/README.md
```diff
@@ -0,0 +1,67 @@
...
+This function displays a progress bar and prints the current progress percentage and remaining computation time.
+Several flags are available to customize the progress bar:
+- `progress_bar=True`: displays the progress bar if True.
+- `percent=True`: displays the progress percentage if True.
+- `start=None`: displays elapsed time from start if not None.
```
yomach, line 37: Confusing name, I would call the variable `start_time=none`

qualang_tools/results/results.py
```diff
@@ -68,11 +58,37 @@ def fetch_all(self):
...
+
+
+def progress_counter(iteration, total, progress_bar=True, percent=True, start=None):
+    """Displays progress bar and prints remaining computation time.
+
+    :param iteration: current iteration. Must be a python integer.
+    :param total: total number of iteration. Must be a python integer.
+    :param progress_bar: Flag enabling the progress bar display. Must be True or False. Default is True.
+    :param percent: Flag enabling the progress percentage display. Must be True or False. Default is True.
+    :param start: Starting time of the program enabling the elapsed time display. Must be the result of time.time(). Default is None.
```
yomach, line 78: `start_time`

qualang_tools/results/results.py
```diff
@@ -68,11 +58,37 @@ def fetch_all(self):
...
+    :return: None.
+    """
+    current_percent = (iteration + 1) / total * 100
+
+    progress = "Progress: "
+    if progress_bar:
+        bar = "#" * int(current_percent / 2)
+        progress += "[" + bar + " " * (50 - len(bar)) + "] "
+    if percent:
+        progress += f"{current_percent:.1f}%"
```
yomach, line 88: If we're rounding it to integers, why display a decimal point?

qualang_tools/results/results.py
```diff
@@ -68,11 +58,37 @@ def fetch_all(self):
...
+    if start is not None:
+        progress += f"--> elapsed time: {time.time()-start:.2f}s"
+
+    print(progress, end="\r")
+    if current_percent == 100:
```
yomach, line 93: This will work poorly if someone stops a run in the middle...
Two things:
1. At the beginning have `print(end="\r")` and at the end (here) have `print(progress, end="")`. This would at least make it end at the end of the line, WDYT?
2. Getting a new line when the use interupts is hard... Maybe we need to add an empty print to the "on_close"?

qualang_tools/units/README.md
```diff
@@ -0,0 +1,42 @@
+# Units tools
```
yomach, line 1: Very nice

qualang_tools/units/units.py
```diff
@@ -0,0 +1,61 @@
...
+        self.GHz = 1000_000_000
+
+        # Voltage units
+        self.V = 1
+        self.mV = 1e-3
+        self.uV = 1e-6
+
+    # conversion functions
+    def to_clock_cycles(self, t):
+        """Converts a duration to clock cycles.
```
yomach, line 34: Need to note it'll be rounded down.

yomach: And / or print a warning

qualang_tools/units/units.py
```diff
@@ -0,0 +1,61 @@
...
+        :return: duration in clock cycles.
+        """
+        if not float(t).is_integer():
+            print(
+                f"Warning: the specified duration ({t}) to be converted to clock cycles in not an integer. It has been converted to int ({int(t)}) to avoid subsequent errors."
+            )
+            t = int(t)
+        return t // 4
+
+    def demod2volts(self, data, duration):
```
yomach, line 46: Ain't there another factor of sqrt(2) or something?
Need to verify with a real OPX.

qualang_tools/loops/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```python
+from qualang_tools.loops import from_array
+import numpy as np 
+
+pulse_amplitude = np.arange(-0.3, 0.3 , 0.01)
+
+with program() as prog:
+    a = declare(fixed)
+    with for_(*from_array(a, pulse_amplitude, "fixed")):
```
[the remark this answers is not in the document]
TheoLaudatQM: I do too, please keep me posted !

qualang_tools/results/results.py
TheoLaudatQM: I don't think I am rounding it...

qualang_tools/results/results.py
TheoLaudatQM: I tested it and it seems that the KeyboardInterrupt  call automatically goes to a new line in the console so I don't think there will be an issue here.
I tested your solution and the problem is that when the loop ends correctly, you do not start on a new line on the console which is what I wanted...

qualang_tools/units/units.py
TheoLaudatQM: I added it to the docstring

qualang_tools/units/units.py
TheoLaudatQM: I am not aware of that and I don't think it is on the documentation so please keep me updated

qualang_tools/loops/loops.py
```diff
@@ -0,0 +1,205 @@
...
+                )
+            else:
+                return (
+                    var,
+                    float(start),
+                    var > float(stop) + float(step) / 2,
+                    var + float(step),
+                )
+    # Logarithmic increment
+    elif increment == "log":
```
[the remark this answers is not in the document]
yomach: I see.
In any case, if it is not supported it should be noted.
Wait, maybe we can do it using "Cast.mult_int_by_fixed"?
For this we support floats <8 or integers above 8.

qualang_tools/results/results.py
yomach: If you tested it and it seems to work, then great :)

qualang_tools/units/units.py
yomach: I checked it and it seems you are right.
In the documentation, we have an extra factor of 2 which is not correct (but we do not know why).
Your function is good.

qualang_tools/loops/loops.py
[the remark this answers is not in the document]
TheoLaudatQM: If I understand well, I think that this multiplication should be made in QUA dsl and not py-qua-tools because in occurs within the QUA for loop, right?
Maybe for now I can just specify that it only works for fixed QUA variables and we'll think of a way of making it work with integer...

qualang_tools/loops/README.md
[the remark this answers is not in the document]
TheoLaudatQM: Do you have any news regarding this point ?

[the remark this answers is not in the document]
TheoLaudatQM: I now understand and tested what we discussed on Slack, so this is good!

qualang_tools/loops/loops.py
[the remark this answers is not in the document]
yomach: I'm not sure I follow.
We don't have implicit casting in QUA, which means that:
`assign(a, b*c)` would fail if `a`, `b` and `c` are not of the same type.
This is how it is defined in the DSL, and this we will not change.
Doing these multiplications needs to be explicit using the `Cast` class.
In this module, If someone wants to have a log scale for an integer variable with a fixed variable step, we can easily support it behind th scenes. And this is a common use-case, so I think we should.
We can also support the other way around, (fixed variable, int increase), but that's much less common.

qualang_tools/loops/loops.py
```diff
@@ -32,9 +32,25 @@ def from_array(var, array, var_type):
         increment = "log"
     else:
         raise Exception(
-            "The spacing of the input array must be linear or logarithmic. Please use `for_each_()` for arbitrary scans."
+            "The spacing of the input array must be even in linear or logarithmic scales Please use `for_each_()` for arbitrary scans."
```
yomach, line 35: Add a "."

qualang_tools/loops/loops.py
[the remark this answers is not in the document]
TheoLaudatQM: Yes, I understand, but there is still a significant rounding issue because `Cast.mul_int_by_fixed()` will convert each point to an integer which will be smaller than the real value and since it is this integer that will be multiplied by the step to get the next one, the error will increase with the number of iterations. A better way would be to keep the fixed point value for the calculation of the next sample and only return the integer.
It is not easy for me to explain like this but for instance, what I am doing now is `item1 = int(item0 * step)`,` item2 = int(item1 * step) = int(int(intem1*step) * step) < int(item1 * step)` and what we need is `item2 = int(item1 * step)`
The bigger the array, the bigger the error...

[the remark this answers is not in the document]
TheoLaudatQM: Yes ok I understand, what bothers me is that the QUA array will not match the corresponding python array so it can be confusing for the user if used for calibration purposes...
I can print a warning explaining it, but do you think that `itemn = int(item0 * step**n)` will be possible one day?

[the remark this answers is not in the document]
yomach: Manually in a while loop that is easy (as long as step**n is < 8), but in a for loop it is problematic...

Ok. the way I see it there are two cases:
1. In the `from_array` function: Here it is very simple, we can manually check if the values are different or not. If they are equal, no problem. If they are different, this is the "worst" thing, as the user gave an array and got something which is slightly different, so we **definitely** need to output a warning message.
2. In the `logspace` function. This is similar to the above. As comping against `np.logspace` might give a different result... So maybe here we should also output a warning?

Maybe we can also add a function that returns the real array?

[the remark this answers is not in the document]
TheoLaudatQM: Yeah, the thing is that the two arrays will almost always be different with logarithmic increment and QUA integers as soon as you have more than 5 points...
I'll put the warnings and create another function to retrieve the exact values taken by the QUA variable

qualang_tools/loops/README.md
```diff
@@ -19,11 +20,31 @@ pulse_amplitude = np.arange(-0.3, 0.3 , 0.01)
...
+exact_qubit_frequency = get_equivalent_log_array(qubit_frequency)
+
+with program() as prog:
+    f = declare(int)
+    with for_(*from_array(f, qubit_frequency)):
+        # The variable 'f' will be looped over the values from exact_qubit_frequency
+        ...
+```
+
+Note that this does not happen when looping over a QUA variable of type fised, or when using a linear increment.
```
yomach, line 46: "fixed"

qualang_tools/loops/loops.py
```diff
@@ -7,23 +7,25 @@
...
     """
+
     # Check for array length
     if len(array) == 0:
         raise Exception("The array must be of length > 0.")
     # Check QUA vs python variables
     if not hasattr(var, "_exp"):
         raise Exception("The first argument must be a QUA variable.")
-    if hasattr(array[0], "_exp"):
+    if not isinstance(array[0], (np.generic, int, float)):
```
yomach, line 28: I guess this is ok?
Make sure you test it with all kind of various inputs... Including a np.array (of a np.array!)

qualang_tools/loops/loops.py
```diff
@@ -199,7 +245,78 @@ def qua_logspace(var, start, stop, num):
...
+    a_log = []
+    aprev = int(log_array[0])
+    step = np.mean(np.array(log_array[1:]) / np.array(log_array[:-1]))
+    for k in range(len(log_array)):
+        a_log.append(aprev)
+        aprev = int(aprev * step)
+    return np.array(a_log)
+
+
+def get_qua_variable_type(variable):
```
yomach, line 309: add underscore:
`_get_qua_variable_type`

This is a private function

yomach (commented):
I think we're almost ready to publish this!!!

qualang_tools/loops/loops.py
TheoLaudatQM: Yes, I tested it and added the case where it is a bool for good measure...

TheoLaudatQM:
Yes, finally !
One point though is that the black version specified in the project (poetry.lock and pyproject.toml) is not the latest and this is the reason why the formatting test fails... Do you know how to fix this? Can I just modify those files manually ?

yomach (approved):

yomach:
I guess we can just merge. Unless there are any other comments?
@qguyk ?

qguyk:
looks good :) we now have a progress bar!

PR #105, opened 2022-06-21
added flattop gaussian

yomach:


yomach:
@KevinAVR @TheoLaudatQM not sure about the definition of the "sigma" in all of the cases, need to think about it!

qualang_tools/config/waveform_tools.py
```diff
@@ -83,3 +87,90 @@ def drag_cosine_pulse_waveforms(amplitude, length, alpha, delta, detuning=0):
...
+):
+    """
+    Returns a flat top cosine waveform. This is a square pulse with a rise and fall with cosine shape with the given
+    sigma. It is possible to only get the rising or falling parts, which allows scanning the flat part length from QUA.
+    The length of the pulse will be the `flat_length + 2 * rise_fall_length`.
+
+    :param float amplitude: The amplitude in volts.
+    :param int flat_length: The flat part length in ns.
+    :param float rise_fall_length: The rise and fall times in ns, taken as a number of points of a cosine between 0
+        and pi.
```
TheoLaudatQM, line 134: I would say "taken as the time for a cosine to go from 0 to 1 (pi phase-shift) and conversely."

qualang_tools/config/waveform_tools.py
```diff
@@ -83,3 +87,90 @@ def drag_cosine_pulse_waveforms(amplitude, length, alpha, delta, detuning=0):
...
+def flattop_tanh_waveform(amplitude, flat_length, rise_fall_length, return_part="all"):
+    """
+    Returns a flat top tanh waveform. This is a square pulse with a rise and fall with tanh shape with the given
+    sigma. It is possible to only get the rising or falling parts, which allows scanning the flat part length from QUA.
+    The length of the pulse will be the `flat_length + 2 * rise_fall_length`.
+
+    :param float amplitude: The amplitude in volts.
+    :param int flat_length: The flat part length in ns.
+    :param float rise_fall_length: The rise and fall times in ns, taken as a number of points of a tanh between -4
+        and 4.
```
TheoLaudatQM, line 161: I would say "taken as the time for a tanh to go from -1 to 1 and conversely."

yomach: But we don't reach exactly -1 and 1, as this would be infinite...

yomach:
> Nice, note that the First and last points of the Gaussian and tanh function are not strictly 0, is it okay or would it be better to substract the offset ?

I wanted to make these functions simpler, I don't think that too many people use the subtraction and the values are very very close to 0.
We can always add it later with an optional parameter... WDYT?

qualang_tools/config/waveform_tools.py
KevinAVR: I'm on Yoav's side. Also probably we can rely on people's knowledge of tanh that it does not reach fixed values.

TheoLaudatQM (approved):

TheoLaudatQM (approved):

PR #113, opened 2022-07-01
Calibration API

TheoLaudatQM:
calibration_API.py contains an API to run several single qubit calibration sequences from one simple python file called calibrations_test.py. This test file will be removed before merging but I let it here to test the API conveniently.


qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
+# Calibration API
+
+## Introduction
+
+This module allows to easily perform most of the standard single qubit calibrations from a single python file.
+
+The following calibrations are currently available:
+* [raw_traces](#raw-traces): acquires the raw and averaged traces from the two inputs of the OPX.
+* [time_of_flight](#time-of-flight): measures the time of flight and offsets of the two inputs of the OPX.
+* [resonator_spectroscopy](#resonator-spectroscopy): performs the resonator spectroscopy.
```
yomach, line 10: **a** resonator spectroscopy

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+* [Rabi](#rabi-measurement): performs a Rabi experiment where the pulse amplitude, duration and frequency can be scanned.
+* [T1](#t1-measurement): measures the qubit T1 by scanning the time after the application of a $\pi/2$ pulse.
+* [Ramsey](#ramsey-measurement): performs a Ramsey measurement where the idle time and pulse frequency can be scanned.
+
+## Usage Example
+1. First the user needs to specify the relevant configuration parameters that will be used for the calibrations.
+Are resuired the _qubit_ and _readout-resonator_ elements and operations, the _integration weights_ and OPX inputs (defined as elements outputs).
```
yomach, line 17: Required variables: ....
Optional variables: ...

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+An optional _flux_line_ element and operation can be specified in the case of flux tunable transmon to find the flus soft spot with the resonator spectroscopy.
+Several options are also available for customizing the plots. All of them are written in the snippet below.
+
+2. Then the calibration API is initialized with these parameters. The standard configuration also needs to be provided.
```
yomach, line 21: A matching(?) configuration also needs to be provided.

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+3. The sequence of calibrations needs to be set using the `.set_` method. All the avainlable calibrations and setting methods are detailed in the *Available calibrations* section below. Note that several calibrations can be set and executed sequentially.
```
yomach, line 22: available

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+4. One needs to open a quantum machine manager with the correct IP address.
```
yomach, line 23: You're changing the tense between the lines. Sometimes it's indirect: "the sequence of calibrations needs to be set" and sometimes it's direct: "One needs to open a quantum machine manager".
I would change it to: "A quantum machine**s** manager needs to be opened"

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+)
+
+# Set calibrations from the ones detailed below
+my_calib.set_...
+
+# Open quantum machines manager
+qmm = QuantumMachinesManager(host="127.0.0.1", port=80)
+
+# Simulate the calibrations
+my_calib.simulate_calibrations(qmm, simulation_duration=5000)
```
yomach, line 73: I would make it such that the `simulate_calibration` can accept both `qmm` and `qm`

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+## Available calibrations
+### Raw traces
+This program acquires the raw and averaged traces from the two inputs of the OPX.
+It only requires two inputs:
+* `iterations`: number of averaging iterations. Must be a positive Python integer.
+* `cooldown_time`: wait time between two iterations in clock cycle unit (4ns). Must be an integer.
+```python
+my_calib.set_raw_traces(iterations=1000, cooldown_time=0)
```
yomach, line 87: Can you put 0? or <4 cycles?
Need to change the example and maybe need to assert on that?

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+```
+![raw_traces](raw_traces.PNG)
+
+### Time of flight
+This program measures the time of flight and offsets of the two inputs of the OPX.
+It requires three inputs:
+* `iterations`: number of averaging iterations. Must be a positive Python integer.
+* `cooldown_time`: wait time between two iterations in clock cycle unit (4ns). Must be an integer.
+* `threshold`: detection threshold for time of flight estimation. The default value is 10/4096 which is usually good enough to detect the waveform.
```
yomach, line 96: In what units is this? =

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+```python
+my_calib.set_time_of_flight(iterations=1000, cooldown_time=0, threshold=10/4096)
+```
+![time_of_flight](time_of_flight.PNG)
+
+### Resonator spectroscopy
+This program measures the resonator response for different readout pulse frequencies defined in scan_variables.
```
yomach, line 103: Maybe write which parameters can be scanned?

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,167 @@
...
+scans = [("frequency", np.arange(100e6, 300e6, step=0.1e6))]
+my_calib.set_resonator_spectroscopy(scan_variables=scans, iterations=1000, cooldown_time=10_000//4)
+```
+
+### Rabi measurement
+This program performs a Rabi-like measurement by sending a pulse to the qubit and measuring the resonator response.
+
+It requires three inputs:
+* `scan_variables`: variables to be scanned. Must be an array of tuples [('variable', [values])]. 1D and 2D scans are available and the scanning parameters are:
+  * 'duration': pulse length in clock cycles (4ns), must be an integer.
```
yomach, line 119: If the pulse is arbitrary then the duration must be longer than it. Maybe we can also assert that?

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
+"""
```
yomach, line 1: I would remove the _API from the name, just "calibrations"
WDYT?

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy import signal
+
+
+available_variables = ["frequency", "amplitude", "duration"]
+u = unit()
+
+
+class QUA_calibrations:
```
yomach, line 19: Need docstring for class

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
...
+        """The QUA calibrations class. It has an API for performing various basic resonator and qubit calibrations.
+        Currently the available calibrations are: 'resonator_spectroscopy', 'rabi' and 'ramsey'.
+        All these calibrations are implemented with 1D or 2D scans among three variables ('frequency', 'amplitude', 'duration').
+        The meaning of these variables is calibration specific.
+        For instance 'duration' represent the qubit pulse length for Rabi and the idle time for Ramsey.
+        Each calibration can be set using the `set_` (`set_rabi` for instance) method and executed using the
+        `run_calibrations()` method or simulated using the `simulate_calibrations` method.
+        The results are stored in a python dictionary called `results`.
+
+        :param configuration:
```
yomach, line 362: missing

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
...
+        """Sets the resonator spectroscopy calibration.
+        This program will measure the resonator response for different readout pulse frequencies defined in scan_variables.
+
+        :param scan_variables: variables to be scanned. Must be an array of tuples [('variable', [values])]. 1D and 2D scans are available.
+        :param iterations: number of averaging iterations. Must be a positive Python integer.
+        :param cooldown_time: resonator or qubit cooldown time in clock cycles (4ns). Must be an integer.
+        :return: None
+        """
+        calibration = "Resonator_spec"
+        self._user_checks(calibration, scan_variables, iterations, cooldown_time)
```
yomach, line 418: The function `_user_checks` sounds like it is only checking things, but in fact, it also adds the calibration to the list.
I would separate the two:
```
self._user_checks()
self._list.append()
```

yomach: Also, you currently start with adding the calibration to the list, which means it'll be added even if the parameters are invalid.
It's better to first do the validation and then add it.
You can also just rename the function to `_add_calibration_to_list()` and just change the order inside.

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
...
+
+        :param scan_variables: None
+        :param iterations: number of averaging iterations. Must be a positive Python integer.
+        :param cooldown_time: resonator or qubit cooldown time in clock cycles (4ns). Must be an integer.
+        :param threshold: detection threshold for time of flight estimation.
+        :return: None
+        """
+        calibration = "time_of_flight"
+        self._user_checks(calibration, scan_variables, iterations, cooldown_time)
+        self.user_specifics["threshold"] = threshold
```
yomach, line 500: shouldn't this be part of the specific calibration?

qualang_tools/addons/calibration/calibrations_API.py
```diff
@@ -0,0 +1,1114 @@
...
+        for key in plot_options:
+            if key in self.plot_options.keys():
+                self.plot_options[key] = plot_options[key]
+            else:
+                raise Exception(
+                    f"Plot option {key} is not implemented, available options are {self.plot_options.keys()}"
+                )
+
+    def _user_checks(self, calibration, scan_variables, iterations, cooldown_time):
+        self.calibration_list.append(calibration)
```
yomach, line 69: What happens if a user adds the same calibration twice? I think it would break things... It would be added to the list twice, but the variables would be overridden and only used once.

yomach:
Looks like a very nice class!

qualang_tools/addons/calibration/README.md
TheoLaudatQM: I thought that best practice is to use simulate with qmm, if I run it with qm I get a warning saying that I should use qmm...

qualang_tools/addons/calibration/README.md
TheoLaudatQM: Yes, if you put 0 then it doesn't wait, if you put >0 and < 4 cc then you have an exception and if you put > 4 it works, assertion is already done

qualang_tools/addons/calibration/README.md
TheoLaudatQM: Yes good point

qualang_tools/addons/calibration/README.md
TheoLaudatQM: it should be something like V/ns, I ass it

qualang_tools/addons/calibration/README.md
TheoLaudatQM: Very good point to avoid what happened at IST...

qualang_tools/addons/calibration/calibrations_API.py
TheoLaudatQM: I don't know because this way I can directly append the variables to scan_var and results if it is valid. Otherwise I would have to duplicate the same loops into two separated functions... Do you think it is necessary or can I just rename the function to _user_checks_and_initialization()?

qualang_tools/addons/calibration/calibrations_API.py
TheoLaudatQM: It is, it is only part of the time of flight calibration. I don't understand

qualang_tools/addons/calibration/calibrations_API.py
TheoLaudatQM: Correct, I didn't think of this case...

qualang_tools/addons/calibration/README.md
yomach: You're right, but since people are still using it, I would support both ways (just to make it easier for customers).

I also think we should remove the deprecated message, but this is for another topic

qualang_tools/addons/calibration/calibrations_API.py
yomach: Not sure I follow, let's discuss

qualang_tools/addons/calibration/calibrations_API.py
yomach: I think it should be: `self.user_specifics["time_of_flight"]["threshold"]`

qualang_tools/addons/calibration/README.md
TheoLaudatQM: We can discuss it, but if best practice is to use qmm (I don't know why actually), then we should update codes on the github that follow these best practices... Especially here, since they only have to write a single line of code `my_calib.simulate_calibrations(qmm, simulation_duration=5000)`

qualang_tools/addons/calibration/calibrations_API.py
TheoLaudatQM: Oh I see, yes !

qualang_tools/addons/calibration/calibrations_API.py
TheoLaudatQM: Now I loop over the variables and if it is valid I update the corresponding dictionaries. If I split into two functions, I have to loop over them first for the check and a second time to update the dictionaries in the initialization function, but you're right it may be cleaner, I was just lazy...

qualang_tools/addons/calibration/README.md
yomach: I don't think there is a need to create two simulation functions, we can just make simulate_calibrations accept both a `QuantumMachinesManager` object and a `QuantumMachine` object. 
First line in the function can be "`if isinstance(machine, QuantumMachinesManager)`

We can also ditch this idea, it's not that critical, just thought it'll be nice.

TheoLaudatQM: Ok done :)

qualang_tools/addons/calibration/README.md
```diff
@@ -7,20 +7,20 @@ This module allows to easily perform most of the standard single qubit calibrati
...
-* [resonator_spectroscopy](#resonator-spectroscopy): performs the resonator spectroscopy.
+* [resonator_spectroscopy](#resonator-spectroscopy): performs a resonator spectroscopy.
 * [Rabi](#rabi-measurement): performs a Rabi experiment where the pulse amplitude, duration and frequency can be scanned.
 * [T1](#t1-measurement): measures the qubit T1 by scanning the time after the application of a $\pi/2$ pulse.
 * [Ramsey](#ramsey-measurement): performs a Ramsey measurement where the idle time and pulse frequency can be scanned.
 
 ## Usage Example
 1. First the user needs to specify the relevant configuration parameters that will be used for the calibrations.
-Are resuired the _qubit_ and _readout-resonator_ elements and operations, the _integration weights_ and OPX inputs (defined as elements outputs).
+Are required the _qubit_ and _readout-resonator_ elements and operations, the _integration weights_ and OPX inputs (defined as elements outputs).
```
yomach, line 17: Need to rephrase, "are required" is not a way to start a sentence.

qualang_tools/addons/calibration/calibrations.py
```diff
@@ -0,0 +1,1203 @@
...
+
+    def set_time_of_flight(
+        self, scan_variables=(), iterations=1, cooldown_time=None, threshold=10 / 4096
+    ):
+        """Measures the time of flight and offsets of the two inputs of the OPX.
+
+        :param scan_variables: None
+        :param iterations: number of averaging iterations. Must be a positive Python integer.
+        :param cooldown_time: resonator or qubit cooldown time in clock cycles (4ns). Must be an integer.
+        :param threshold: detection threshold for time of flight estimation.
```
yomach, line 549: Add units here as well

yomach (commented):
Looks good. Gave a few last small comments.
Probably best to play around with it a bit more, as I only reviewed from the code side...

yomach (approved):

TheoLaudatQM:
Thanks for the comments, indeed I tested it a lot already but I may have missed something... We don't have to merge it now, but I would like to have it on the github before the next installation in order to test it on a real device

qualang_tools/addons/calibration/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+This module allows to easily perform most of the standard single qubit calibrations from a single python file.
+
+The following calibrations are currently available:
+* [raw_traces](#raw-traces): acquires the raw and averaged traces from the two inputs of the OPX.
+* [time_of_flight](#time-of-flight): measures the time of flight and offsets of the two inputs of the OPX.
+* [resonator_spectroscopy](#resonator-spectroscopy): performs a resonator spectroscopy.
+* [Rabi](#rabi-measurement): performs a Rabi experiment where the pulse amplitude, duration and frequency can be scanned.
+* [T1](#t1-measurement): measures the qubit T1 by scanning the time after the application of a $\pi/2$ pulse.
+* [Ramsey](#ramsey-measurement): performs a Ramsey measurement where the idle time and pulse frequency can be scanned.
```
KevinAVR, line 13: it would be nice to add Ramsey with `frame_rotation_2pi()` this is for the case wehre you want to detune like 10 MHz and still see good oscillations with the previously calibrated pi pulse. Gives accurate results as well

qualang_tools/addons/calibration/calibrations.py
```diff
@@ -0,0 +1,1208 @@
...
+    def _align_measure_save(self, I, Q, I_st, Q_st, current_calib, align_flag=True):
+        if align_flag:
+            align(self.qubit_elmt, self.readout_elmt)
+        # Measure the resonator
+        if len(self.out) == 2:
+            measure(
+                self.readout_op,
+                self.readout_elmt,
+                None,
+                dual_demod.full(self.iw[0], self.out[0], self.iw[1], self.out[1], I),
```
KevinAVR, line 196: Can this be used when the upconversion does LO - IF? then the integration weights need to be flipped around. or do we assume that we would swap the cables in the OPX or in the config?

qualang_tools/results/README.md
```diff
@@ -6,20 +6,20 @@ The fetching tool has an API to easily fetch data from the stream processing.
...
+- `mode="live"` will fetch data one by one for all named results for live plotting purposes.
 
 ### Usage example
 
  
 ```python
-from qualang_tools.results import result_tool
+from qualang_tools.results import fetching_tool
 
 qmm = QuantumMachinesManager()
```
KevinAVR, line 18: do you need to specify the IP in the example?

qualang_tools/addons/calibration/calibrations_test.py
```diff
@@ -0,0 +1,228 @@
...
+)
+
+# Raw traces
+# my_calib.set_raw_traces(iterations=n_avg, cooldown_time=0)
+
+# Time of flight
+# my_calib.set_time_of_flight(iterations=n_avg, cooldown_time=0)
+
+# Open QM
+qmm = QuantumMachinesManager(host="172.16.2.103", port=80)
```
KevinAVR, line 216: in the test you need to be connected thru the VPN?

qualang_tools/addons/calibration/calibrations.py
```diff
@@ -0,0 +1,1208 @@
...
+                                )
+                            # Second pi/2 pulse
+                            play(self.qubit_op, self.qubit_elmt)
+
+                            # Align, measure and save data
+                            self._align_measure_save(I, Q, I_st, Q_st, calib)
+                    save(n, n_st)
+
+                with stream_processing():
+                    I_st.buffer(len(var2_scan), len(var1_scan)).average().save("I")
```
KevinAVR, line 978: maybe we can change it to `I_st.buffer(len(var1_scan)).buffer(len(var2_scan))` this is more intutitive for me

KevinAVR (commented):
minor comments seems cool :)

qualang_tools/addons/calibration/calibrations.py
yomach: If you put a positive IF and you get LO-IF, then you need to flip things. I think that this is the best practice and we should strongly encourage our customers to do so.
It could also be that you output LO+IF but you still need to flip the integration weights, this is if the down conversation cables are swapped... In this case, we also need to flip it (config or physically).
It also drastically simplifies the code, and specifically things like the discriminators and weight optimizers.

qualang_tools/addons/calibration/README.md
TheoLaudatQM: Do you mean add a constant phase-shift for the 2nd pi/2 pulse or get the oscillations by scanning the phase-shift of the 2nd pi/2 pulse?

qualang_tools/addons/calibration/calibrations.py
TheoLaudatQM: I agree with what Yoav said and here you can anyway just swap the integration weights that you input to the function when you declare the calibration API (cf line 164 of calibration_test.py) so it should not change this function

qualang_tools/results/README.md
TheoLaudatQM: Right I should add it for good measure

qualang_tools/addons/calibration/calibrations_test.py
TheoLaudatQM: This is how I tested it yes

qualang_tools/addons/calibration/calibrations.py
TheoLaudatQM: Alright, the syntax is `I_st.buffer(1st_inner_loop).buffer(2nd_inner_loop).buffer(3rd_inner_loop)...` right?

PR #121, opened 2022-07-20
update test and waveform tool

KevinAVR:
I was missing a 2*pi in the AC detuning


qualang_tools/config/waveform_tools.py
```diff
@@ -42,7 +42,7 @@ def drag_gaussian_pulse_waveforms(
     z = gauss_wave + 1j * 0
     if alpha != 0:
         # The complex DRAG envelope:
-        z += 1j * gauss_der_wave * (alpha / (delta - detuning))
+        z += 1j * gauss_der_wave * (alpha / (delta - 2 * np.pi * detuning))
```
TheoLaudatQM, line 45: I guess the 2pi should be in factor of (delta-detuning) because both delta and detuning are in Hz, right ?
Also I think we should replace detuning by anharmonicity whose meaning is clearer and closer to the referenced thesis and paper, what do you think?

tests/test_waveform_tools.py
```diff
@@ -110,10 +110,10 @@ def test_drag_detune():
...
...001681587624964176, 0.002979222690258811, 0.004634421599986756, 0.006637311388457736, 0.008975848335524833, 0.011635835120196633, 0.014600943905999777, 0.0178527466431549, 0.02137075398014205, 0.02513246424921537, 0.02911342402414721, 0.033287301741886516, 0.03762597583158239, 0.04209963870397941, 0.04667691782174653, 0.0513250148978447, 0.05600986405633606, 0.06069630954061614, 0.06534830327116405, 0.06992912224249428, 0.07440160541163188, 0.07872840937326084, 0.08287228174535365, 0.08679635080962321, 0.09046442956992089, 0.09384133201533117, 0.096893199009905, 0.09958783088447047, 0.101895023484411, 0.10378690413716896, 0.1052382637506687, 0.10622688104461486, 0.10673383475595741, 0.10674379955238371, 0.10624532133747262, 0.10523106764134782, 0.10369804886369208, 0.10164780627334541, 0.09908656287100143, 0.09602533348837639, 0.0924799908273187, 0.08847128453334618, 0.08402481084676613, 0.07917093087663929, 0.07394463609327032, 0.06838536022768293, 0.06253673739491354, 0.056446306914480576, 0.0501651659779772, 0.04374757200181161, 0.03725049719368061, 0.03073313854511192, 0.024256387129911346, 0.01788226123008625, 0.011673308417393868, 0.005691982280900334, 1.2711786904350342e-17]
```
TheoLaudatQM, line 113: Here you just updated the test waveforms based on the 2pi factor, right?

TheoLaudatQM (changes requested):

TheoLaudatQM (changes requested):
We just need to clarify this 2pi factor

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+### Fixed
+- **Breaking change!** - Waveform tools - Added a missing 2pi factor into `detuning` parameter in `drag_gaussian_pulse_waveforms` and `drag_cosine_pulse_waveforms`. 
+  This will produce different results compared to previous versions, to get the same results, divide the `detuning` parameter by 2pi.
```
yomach, line 9: This now needs to be changed also for the anharmonicity.
But I'm not sure about this change, why not just change the docstring to indicate that we expect it in Hz (and not in Rad)?

KevinAVR: Thanks, I updated.

KevinAVR: The docstrings mention already that the units should be in Hz

yomach: I mean:
This is a breaking change, anyone updating qualang tools will have their script not working and will wonder why.
99.9% of the people do not look at the changelog... Did you ever look at the changelog when you updated Numpy? Scipy? qm-qua?? :) 
How important is it to have this change? Can't we update the docstring **instead** of doing this change?
Options we have:
1. Keep the breaking change as it is now, I don't really like this option.
2. Revert this breaking change completely, updating the docstring
3. We can use the changed variable name: If the user-supplied `delta` it will be in Rad/s and if the user gave `Anharmonicity` then it will be in Hz. But this means we need to change the function call a little bit and we also need to change the name of the detuning parameter...

WDYT? @TheoLaudatQM @KevinAVR 

I was wrong in my previous comment, I meant to write that we expect it in Rad/s and not in Hz.

KevinAVR: I like the third option.

KevinAVR: I don't follow why you need to change the name of the detuning parameter

yomach: I'll elaborate more with a code example tomorrow

yomach: I'm commenting outside of the "conversation" for easier finding later

yomach:
This is the original function definition:
```python
def drag_gaussian_pulse_waveforms(amplitude, length, sigma, alpha, delta, detuning=0, subtracted=True):
```
Most customers simply access the variables by location, but they can also do it with the var name, any of these are equivalent:
```python
drag_gaussian_pulse_waveforms(0.1, 40, 5, 0.2, 5e6, 1e6)
drag_gaussian_pulse_waveforms(0.1, 40, 5, 0.2, delta=5e6, detuning=1e6)
drag_gaussian_pulse_waveforms(0.1, 40, 5, 0.2, detuning=1e6, delta=5e6)
```
We want to make sure that all of these work after the change. If we simply rename delta, this would be enough:
```python
def drag_gaussian_pulse_waveforms(amplitude, length, sigma, alpha, anharmonicity, detuning=0, subtracted=True):
    delta = kwargs.get("delta", None)
    if delta is not None:
        anharmonicity = delta
        print("'delta' has been replaced by 'anharmonicity' and will be deprecated in the future.")
```
This would cover **all cases** above.
However, if we also want to add a factor of two pi, we would need to do this:
```python
def drag_gaussian_pulse_waveforms(amplitude, length, sigma, alpha, delta=None, detuning=0, subtracted=True, anharmonicity):
    if delta is not None:
        anharmonicity = delta / 2*pi
        print("'delta' has been replaced by 'anharmonicity' and will be deprecated in the future. Note that anharmonicity = delta / 2*pi")
```
This is because we want to make sure that the first example I showed above would also behave the same `drag_gaussian_pulse_waveforms(0.1, 40, 5, 0.2, 5e6, 1e6)`

Now imagine we also change detuning, we would need a new variable `detuning_new` and do the same trick.

Going back to the three options I wrote above:
1. Keep the breaking change as it is now, I don't really like this option, as it would hurt customers.
2. Revert this breaking change completely, and just update the docstring that we accept Rad/s (I prefer this option).
3. We can use the changed variable name: If the user-supplied delta it will be in Rad/s and if the user gave Anharmonicity then it will be in Hz. But this is pretty complicated and ugly, as you can see above.

PR #123, opened 2022-07-21
Compatibility config schema 0.3.8 alpha1

SatyaBade12:
Numerous updates to make the ConfigBuilder tool compatible with config schema version 0.3.8 alpha 1

### DIFF
--- a/qualang_tools/config/builder.py
+++ b/qualang_tools/config/builder.py
@@ -72,8 +72,8 @@ class ConfigBuilder:
                         "feedback": port.filter.feedback,
                         "feedforward": port.filter.feedforward,
                     }
-                if port.channel_weights is not None:
-                    port_dict["channel_weights"] = port.channel_weights
+                if port.crosstalk is not None:
+                    port_dict["crosstalk"] = port.crosstalk
                 port_dict["offset"] = port.offset
                 port_dict["shareable"] = port.shareable
             elif isinstance(port, AnalogInputPort):
--- a/qualang_tools/config/builder.py
+++ b/qualang_tools/config/builder.py
@@ -84,7 +84,6 @@ class ConfigBuilder:
                 port_dict["shareable"] = port.shareable
             elif isinstance(port, DigitalOutputPort):
                 port_dict = cont.dict["digital_outputs"][port_info]
-                port_dict["offset"] = port.offset
                 port_dict["shareable"] = port.shareable
             elif isinstance(port, DigitalInputPort):
                 port_dict = cont.dict["digital_inputs"][port_info]
--- a/qualang_tools/config/builder.py
+++ b/qualang_tools/config/builder.py
@@ -62,37 +62,39 @@ class ConfigBuilder:
     def _add_controller(self, cont: Controller):
         self.controllers.append(cont)
         for port in cont.ports:
+            port_info = port.info[1]
             if isinstance(port, AnalogOutputPort):
+                port_dict = cont.dict["analog_outputs"][port_info]
                 if port.delay is not None:
-                    cont.dict["analog_outputs"][port.info[1]]["delay"] = port.delay
+                    port_dict["delay"] = port.delay
                 if port.filter is not None:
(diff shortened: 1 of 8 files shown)


qualang_tools/config/components.py
```diff
@@ -684,14 +736,36 @@ def smearing(self, smearing: int):
...
+        :type sines: List[float]
+        :param durations: duration of the each segment
+        :type durations: List[int]
+        """
+        self._durations = durations
+        self._cosines = cosines
+        self._sines = sines
+        super(PiecewiseConstantIntegrationWeights, self).__init__(
+            name,
+            [(cosine, duration) for (cosine, duration) in zip(cosines, durations)],
```
tomerfqm, line 756: The durations of the cosine weights segments doesn't have to match the duration of the sine weights segments

yomach: Are you sure?

qualang_tools/config/builder.py
```diff
@@ -62,37 +62,39 @@ def ports(self):
...
+                    port_dict["filter"] = {
                         "feedback": port.filter.feedback,
                         "feedforward": port.filter.feedforward,
                     }
                 if port.channel_weights is not None:
-                    cont.dict["analog_outputs"][port.info[1]][
-                        "channel_weights"
-                    ] = port.channel_weights
-                cont.dict["analog_outputs"][port.info[1]]["offset"] = port.offset
+                    port_dict["channel_weights"] = port.channel_weights
```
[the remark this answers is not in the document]
SatyaBade12: was it "crosstalk" in v0.3.8-alpha1?

qualang_tools/config/components.py
tomerfqm: Ummmm nope... I'll check

qualang_tools/config/components.py
```diff
@@ -684,14 +736,36 @@ def smearing(self, smearing: int):
...
     def __init__(self, name: str, cosine: float, sine: float, duration: int):
         """A constant integration weight used in integration and demodulation
         :param name: name for the weights vector
         :type name: str
         :param cosine: value of the cosine vector
         :type cosine: float
-        :param sine: value of the cosine vector
+        :param sine: value of the sine vector
         :type sine: float
         :param duration: duration of the read out pulse
```
TheoLaudatQM, line 770: Should it be ns or clock cycles?

qualang_tools/config/components.py
```diff
@@ -684,14 +736,36 @@ def smearing(self, smearing: int):
...
+class PiecewiseConstantIntegrationWeights(IntegrationWeights):
+    def __init__(self, name: str, cosines: List, sines: List, durations: List):
+        """Piece-wise constant integration weights used in integration and demodulation
+        :param name: name for the weights vector
+        :type name: str
+        :param cosines: values of the cosine vector
+        :type cosines: List[float]
+        :param sines: values of the sine vector
+        :type sines: List[float]
+        :param durations: duration of each segment
```
TheoLaudatQM, line 748: Should the duration be in ns or clock cycles?

TheoLaudatQM (approved):

qualang_tools/config/components.py
TheoLaudatQM: Actually, I think they have to otherwise: `qm - ERROR - CONFIG ERROR in key "integrationWeights.Wc" [Integration weights] : Integration weight Wc sine and cosine size is not equal`

tomerfqm: So I meant that the piecewise segments don't have to match in duration (or number). Of course, the total length must be the same. 

For example, the following is valid:
```
"cosine_weights": {
    "cosine": [(1.0, readout_len/2),(0.1, readout_len/2)],
    "sine": [(0.0, readout_len)],
},
```

tomerfqm (changes requested):

tomerfqm (approved):

PR #125, opened 2022-07-22
Update live plotting conditions from Yoav idea

TheoLaudatQM:
Use bcont and blast in results.is_processing()

### DIFF
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -44,7 +44,6 @@ class fetching_tool:
     def is_processing(self):
         """
         Returns True while the program is processing, and also once after the processing is done. Can be used for live plotting.
-        The time at which the processing starts can be access with `self.get_start_time()` to feed `progress_counter()`.
         **Example**: while my_results.is_processing():
 
         :return: boolean flag which is True while the program is processing, and also once after the processing is done.
@@ -67,7 +66,7 @@ class fetching_tool:
 
     def get_start_time(self):
         """
-        Gets the time at which the processing started.
+        Gets the time at which the processing started. To be used in progress_counter().
 
         :return: float for the time at which the processing started.
         """
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -47,7 +47,7 @@ class fetching_tool:
         The time at which the processing starts can be access with `self.get_start_time()` to feed `progress_counter()`.
         **Example**: while my_results.is_processing():
 
-        :return: boolean flag which is True while the program is processing and once after to plot the last iteration.
+        :return: boolean flag which is True while the program is processing, and also once after the processing is done.
         """
         if self.start_time == 0:
             self._b_cont = self.res_handles.is_processing()
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -44,7 +44,7 @@ class fetching_tool:
     def is_processing(self):
         """
         Returns True while the program is processing. Used for live plotting.
-        The time at which the processing starts is stored in `self.start_time` to feed `progress_counter()`.
+        The time at which the processing starts can be access with `self.get_start_time()` to feed `progress_counter()`.
         **Example**: while my_results.is_processing():
 
         :return: boolean flag which is True while the program is processing and once after to plot the last iteration.
@@ -64,6 +64,14 @@ class fetching_tool:
                 if len(data.dtype) == 1:
                     data = data["value"]
         return data
+
+    def get_start_time(self):
+        """
+        Gets the time at which the processing started.
+
+        :return: float for the time at which the processing started.
+        """
+        return self.start_time
 
     def fetch_all(self):
         """
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -66,9 +66,9 @@ class fetching_tool:
 
     def get_start_time(self):
         """
-        Gets the time at which the processing started. To be used in progress_counter().
+        Gets time at which is_processing() was first called. To be used in progress_counter().
 
-        :return: float for the time at which the processing started.
+        :return: float for the time at which is_processing() was first called.
         """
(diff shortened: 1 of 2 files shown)


qualang_tools/results/results.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, job, data_list, mode="wait_for_all"):
                     self.data_handles[-1].wait_for_values(1)
                 else:
                     raise Warning(f"{data} is not saved in the stream processing.")
+            # Live plotting parameters
+            self.__bcont = False
+            self.__blast = True
```
yomach, line 41: why two underscores? Usually, one is used for internal parameters.

qualang_tools/results/results.py
```diff
@@ -44,7 +48,14 @@ def is_processing(self):
 
         :return: boolean flag which is True while the program is processing.
         """
```
yomach, line 50: We also need to update the docstring.
Maybe we also need to rename the function?

qualang_tools/results/results.py
```diff
@@ -98,7 +109,7 @@ def progress_counter(
     if percent:
         progress += f"{current_percent:.1f}%"
     if start_time is not None:
```
yomach, line 111: Now you also have `self.start_time`.
So maybe if the given `start_time` is None, use the one made internally? 
And of course, write it in the docstring

qualang_tools/results/results.py
TheoLaudatQM: I saw that it is two '_' for making variables private: https://www.tutorialspoint.com/private-variables-in-python

qualang_tools/results/results.py
TheoLaudatQM: I don't know, because it does the same thing as before just more efficiently, why do you want to rename it ?

qualang_tools/results/results.py
TheoLaudatQM: Yes but self.start_time belong to the fetching tool class and not to progress_counter which is an independent function

qualang_tools/results/results.py
yomach: Interesting, I didn't know that. But that page is a bit misleading:
https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name
A single _ is for making them "private" (not intended to be changed/accessed by the user). 
A double __ is for making them "private" (not to be changed by sub-children of the class).
We care about the first :)

qualang_tools/results/results.py
yomach: It's not 100% representative, because it also returns true the first time it is called after it is done processing.
But I'm ok with the name, just need to add it to the docstring :)

qualang_tools/results/results.py
yomach: You're right, my bad.
But then, why do we need `self.start_time` for the fetching tool? Is it only being used to check if it ever ran.
we can use a boolean for that:
In the init:
`self._b_first_time = True`

TheoLaudatQM: This is what I did first, but then I figured that by using start_time, we can also directly get the time at which the processing starts and feed it to the progress counter so that the user does not have to import time and write start_time=time()

qualang_tools/results/results.py
TheoLaudatQM: So what I wanted is that the user cannot access these variables which can only be done with double '_' otherwise you can still get them by typing results._blast

yomach: The user can still get them with double `__`, he just need to type a bit more (`results._fetching_tools_start_time`).

In any case, in Python there is no real hidden/private variables, you mark them with a single `_`. If the user wants, he can get them, but we're ok with that.

yomach: btw, blast is confusing as it is a word in english... I would change to `_b_last`

TheoLaudatQM: Ok I don't mind

qualang_tools/results/results.py
yomach: Oh, so you want the users to have access to it?
In this case, I would add a getter function:
`def get_start_time():`
And make it return that time. It is not good practice to directly access the object's variables.

qualang_tools/results/results.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, job, data_list, mode="wait_for_all"):
     def is_processing(self):
         """
         Returns True while the program is processing. Used for live plotting.
+        The time at which the processing starts is stored in `self.start_time` to feed `progress_counter()`.
```
yomach, line 47: I've just noticed you added this. I'm ok with having this and not a getter function (although I think it's nicer).
But I would rephase to:

The time at which this method is first called is stored in `self.start_time`. It can be used to for `progress_counter()`.

qualang_tools/results/results.py
yomach: See my other comment above :)

qualang_tools/results/results.py
```diff
@@ -36,15 +36,27 @@ def __init__(self, job, data_list, mode="wait_for_all"):
...
                 else:
                     raise Warning(f"{data} is not saved in the stream processing.")
+            # Live plotting parameters
+            self._b_cont = False
+            self._b_last = True
+            self.start_time = 0
 
     def is_processing(self):
         """
         Returns True while the program is processing. Used for live plotting.
```
yomach, line 46: "Returns True while the program is processing, and also once after the processing is done. Can be used for live plotting."

qualang_tools/results/results.py
TheoLaudatQM: I did the get_start_time function which is indeed better

qualang_tools/results/results.py
TheoLaudatQM: done

PR #126, opened 2022-07-24
Fitting class

MichalGQM:
Hi, I added the fitting class to pyqua_tools. 
I put it in the plot library, let me know if you think that it should be in a different place :)


qualang_tools/plot/README.md
```diff
@@ -21,3 +21,72 @@ while job.result_handles.is_processing():
...
+
+## Fitting 
+This tool makes the fitting process much easier! Once you have the data you can:
+
+* Fit the data to a suitable fitting function.
+* Plot the data as well as the fitting function.
+* Extract physical parameters from the fitting.
+* Save the fitting parameters to a json file.
+* Open the json file and save its data to a python dictionary.
+
```
TheoLaudatQM, line 33: I would write somewhere in the readme the list of the available fitting function so that the user knows if the one he/she's looking for is implemented

qualang_tools/plot/Fitting/fitting_functions.py
```diff
@@ -0,0 +1,348 @@
...
+    .. math::
+    f(x) = uncertainty_population * (1 - np.exp(-x * (1/tau))) + amp / 2 * (
+        np.exp(-x * (1/tau))
+        * (initial_offset * 2 + np.cos(2 * np.pi * f * x + phase))
+        )
+
+    for unknown parameters `f`, `phase`, `tau`, `amp`, `uncertainty_population` and `initial_offset`.
+
+    :param x:  The data on the x axis
+    :param y: The data on the y axis
```
TheoLaudatQM, line 18: In what units should these be?

TheoLaudatQM, line 18: It looks like this should be the transition probability to be in the excited state right ?

qualang_tools/plot/Fitting/fitting_functions.py
```diff
@@ -0,0 +1,348 @@
...
+
+    .. math::
+    f(x) = uncertainty_population * (1 - np.exp(-x * (1/tau))) + amp / 2 * (
+        np.exp(-x * (1/tau))
+        * (initial_offset * 2 + np.cos(2 * np.pi * f * x + phase))
+        )
+
+    for unknown parameters `f`, `phase`, `tau`, `amp`, `uncertainty_population` and `initial_offset`.
+
+    :param x:  The data on the x axis
```
TheoLaudatQM, line 17: This should be the dephasing time right?

TheoLaudatQM (commented):
Very cool tool, I tested the Ramsey fit and I could not make it crash :)
I added some suggestions for improvements that we can discuss if they are not clear

qualang_tools/plot/Fitting/fitting_functions.py
```diff
@@ -0,0 +1,348 @@
+from scipy import optimize
+import numpy as np
+
+
+def fit_ramsey(x, y):
+    """
+    Create a fit to Ramsey experiment of the form
```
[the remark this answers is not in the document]
MichalGQM: Oh ok, I thought it is clear by saying Ramsey.. but maybe it's better to add it :)

qualang_tools/plot/README.md
```diff
@@ -21,3 +21,72 @@ while job.result_handles.is_processing():
...
+### Use
+In order to use the fitting class, one should have `from fitting import *` in his/her code.
+ 
+### Usage example
+In the example below you can see five parts:
+1. The measured data - This section simulates the data in order to show the relevant functions.
+2. Fit - This section does the fit to the data.
+3. Plot - This section plots the data and the fitting plot.
+4. Save - This section saves the data to a json file. You are required to enter the file name, so the saved file will be of the form 'data_fit_{file_name}.json` 
+5. Read json - This section reads the json file and prints the saved parameters if the flag `print_params=True`.
```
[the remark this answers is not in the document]
MichalGQM: what do you mean? :)

qualang_tools/plot/Fitting/fitting_functions.py
```diff
@@ -0,0 +1,348 @@
...
+            a[5] * initial_offset / peaks[0] * 2
+            + np.cos(2 * np.pi * a[0] * omega / (2 * np.pi) * x + a[3])
+        )
+    )
+
+    def curve_fit(f, x, y, a0):
+        def opt(x, y, a):
+            return np.sum(np.abs(f(x, a) - y) ** 2)
+
+        out = optimize.minimize(lambda a: opt(x, y, a), a0)
```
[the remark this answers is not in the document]
MichalGQM: You are right, it's important to check it. In order to add this flag I first have to make a test. Do you have any idea checking it?

qualang_tools/plot/README.md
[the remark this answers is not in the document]
TheoLaudatQM: Sorry it formatted it so it is not clear, you need to replace the ' before data_fit_{... by `

qualang_tools/plot/Fitting/fitting_functions.py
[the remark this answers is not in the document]
TheoLaudatQM: The variable out.success will automatically be False if the fit does not converge so one way of testing it would be try to fit a Lorentzian with the Ramsey fitting function for instance

qualang_tools/plot/README.md
MichalGQM: Good idea, I'll add it :)

qualang_tools/plot/Fitting/fitting_functions.py
MichalGQM: In the linear fit it depends I think .. in Ramsey is it seconds? or ns ? as well as spectroscopy, is it Hz or MHz?

MichalGQM: So it is actually the transmission in a.u. no?

qualang_tools/plot/Fitting/fitting_functions.py
MichalGQM: Yes, in seconds or ns?

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,134 @@
+from fitting_functions import *
+import itertools
+import json
+import matplotlib.pyplot as plt
+
+
+class Fit:
```
yomach, line 7: All of these classes should be made static (Talk to me about it)

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,134 @@
...
+        fit_params["y_data"] = y.tolist()
+        fit_params["y_fit"] = fit_func["fit_func"](x).tolist()
+        json_object = json.dumps(fit_params)
+        with open(f"data_fit_{id}.json", "w") as outfile:
+            outfile.write(json_object)
+
+
+class Read:
+    """
+    This class takes care of reading the saved.
```
yomach, line 115: reading the saved data

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def save_params(self, x, y, fit_function, id):
+        """
+        Save the data as a json file
+
+        :param x: The data on the x axis
+        :param y: The data on the y axis
+        :param fit_function: The output of the fit function
+        :param id: The name of the saved data
+        :return: A jason file with the given name, which includes all the data.
```
yomach, line 100: json

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+
+        fit_function = fit_reflection_resonator_spectroscopy(x, y)
+        return fit_function
+
+
+class Plot:
+    """
+    This class takes care of plotting the data and the fitted function.
+    """
```
yomach, line 58: The docstrings needs to be expended

qualang_tools/plot/README.md
```diff
@@ -21,3 +21,72 @@ while job.result_handles.is_processing():
...
+read = Read()
+data = read.read_saved_params(id=file_name, print_params=True)
+
+
+```
+
+Below is the plot of this example
+![fitting_example](example.PNG)
+
+# Possible future extensions
```
yomach, line 88: Not sure if we want to list these, are we committing to adding them?

qualang_tools/plot/__init__.py
```diff
@@ -1,3 +1,16 @@
 from qualang_tools.plot.plot import interrupt_on_close
+from qualang_tools.plot.Fitting.fitting import *
+from qualang_tools.plot.Fitting.fitting_functions import *
```
yomach, line 3: Unless we want the user to have access directly to the `fitting_functions`, this should be removed

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,134 @@
+from fitting_functions import *
+import itertools
+import json
+import matplotlib.pyplot as plt
+
+
+class Fit:
+    """
+    This class takes care of the fitting to the measured data.
+    All the fitting functions are taken from the `Fitting_functions` file.
```
yomach, line 10: In python packages, from the user perspective, there are no "files".
If we want user to have access to the functions directly, then we can just rename it to "module". Otherwise it should be removed.

qualang_tools/plot/Fitting/fitting_functions.py
TheoLaudatQM: Then I would call it dephasing time instead of The data on the x axis and I think that the best is to have it in ns because this is the unit it usually is in the QUA program. Be careful that in this case the frequency will be in GHz so we may want to convert it in MHz

qualang_tools/plot/Fitting/fitting_functions.py
TheoLaudatQM: What do you call transmission? Is it sqrt(I**2+Q**2)?

yomach (changes requested):
We want the user to use the "Fit" class, right? This is from the examples.
If this is the case, it means he will never see documentation and docstring in the `fitting_function.py` file. All of these comments need to be moved to the Fit class.

In addition, I think that it better fits the analysis module.

qualang_tools/plot/Fitting/fitting_functions.py
TheoLaudatQM: For the linear case, I actually it is too general to put units.
For Ramsey I would put the dephasing time in ns and for the Y it depends on what data you expect to be there
For spectroscopy I would put Hz because we usually scan the frequency in Hz unit

qualang_tools/plot/Fitting/fitting_functions.py
```diff
@@ -0,0 +1,348 @@
...
+    :param x:  The data on the x axis
+    :param y: The data on the y axis
+    :return: A dictionary of (fit_func, f, phase, tau, amp, uncertainty_population, initial_offset)
+        fit_func - The fitted function
+        f -  The detuning frequency
+        phase - The phase
+        tau - The decay constant
+        amp - The amplitude
+        uncertainty_population -  The uncertainty in population
+        initial_offset - The initial offset
```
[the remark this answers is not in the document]
MichalGQM: I can change tau to be T2, is it more clear? the uncertainty population is actually the amplitude.. and the initial offset is not zero because of noise I guess... ?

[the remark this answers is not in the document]
MichalGQM: I changed the comments :)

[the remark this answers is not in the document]
TheoLaudatQM: If tau is actually T2, then yes I would call it T2
In general I think it is best to use physics related quantities as fitting parameters so that users know directly what they represent.
It is still not clear to me what uncertainty_population represent... Where does this equation come from?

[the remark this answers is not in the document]
MichalGQM: I will talk to Asaf and update it there :)

qualang_tools/plot/Fitting/fitting_functions.py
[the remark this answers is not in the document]
MichalGQM: Oh, ok.. cool thanks !

qualang_tools/plot/README.md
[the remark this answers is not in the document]
MichalGQM: ok

qualang_tools/plot/Fitting/fitting_functions.py
MichalGQM: Yes, it is sqrt(I+Q)

qualang_tools/plot/README.md
MichalGQM: I am about to work on those 3 statements... unless you think we don't need some of them..

qualang_tools/plot/__init__.py
MichalGQM: Do we want? what are the considerations?

qualang_tools/plot/README.md
yomach: I would not write them in the readme.

qualang_tools/plot/__init__.py
yomach: I think it's a bit redundant to be able to access these functions from two different places... But you've made it that way, so I was mainly asking you

qualang_tools/plot/Fitting/fitting_functions.py
TheoLaudatQM: I guess that if we renormalize the data internally to do the fitting it doesn't really matter, it can be I, Q, sqrt(I^2+Q^2), phase, transition probability...

qualang_tools/plot/__init__.py
MichalGQM: Do you think it is better to put everything on the same script?

yomach: yes

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,418 @@
...
+            x,
+            y,
+            [1, 1, 1, angle0, 1, 1, 1],
+        )
+
+        print(
+            f"f = {popt[0] * omega / (2 * np.pi) * xn[1]}, phase = {popt[3] % (2 * np.pi) * xn[1]}, T2 = {(1 / (decay_guess * popt[1])) * xn[1]}, amp = {peaks[0] * popt[2] * yn[1]}, uncertainty population = {post_decay_mean * popt[4] * yn[1]}, initial offset = {popt[5] * initial_offset * yn[1]}"
+        )
+        out = {
+            "fit_func": lambda x: fit_type(x, popt) * yn[1],
```
TheoLaudatQM, line 169: This line should be `"fit_func": lambda x: fit_type(x / xn[1], popt) * yn[1],`

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,418 @@
...
+
+        popt = curve_fit(
+            fit_type,
+            x,
+            y,
+            [1, 1, 1, angle0, 1, 1, 1],
+        )
+
+        print(
+            f"f = {popt[0] * omega / (2 * np.pi) * xn[1]}, phase = {popt[3] % (2 * np.pi) * xn[1]}, T2 = {(1 / (decay_guess * popt[1])) * xn[1]}, amp = {peaks[0] * popt[2] * yn[1]}, uncertainty population = {post_decay_mean * popt[4] * yn[1]}, initial offset = {popt[5] * initial_offset * yn[1]}"
```
TheoLaudatQM, line 166: The phase shouldn't be rescaled by `xn[1]`

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,418 @@
...
+        )
+        out = {
+            "fit_func": lambda x: fit_type(x, popt) * yn[1],
+            "f": popt[0] * omega / (2 * np.pi) * xn[1],
+            "phase": popt[3] % (2 * np.pi) * xn[1],
```
TheoLaudatQM, line 171: The phase shouldn't be rescaled by `xn[1]`

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,418 @@
...
+        xn = preprocessing.normalize([x], return_norm=True)
+        yn = preprocessing.normalize([y], return_norm=True)
+        x = xn[0][0]
+        y = yn[0][0]
+
+
+        w = np.fft.fft(y)
+        freqs = np.fft.fftfreq(len(x))
+        new_w = w[1: len(freqs // 2)]
+        new_f = freqs[1: len(freqs // 2)]
```
TheoLaudatQM, line 111: I guess that it should be `w[1: len(freqs) // 2]` no?
If so, then you don't need lines 113 to 115

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -0,0 +1,418 @@
...
+        out = {
+            "fit_func": lambda x: fit_type(x, popt) * yn[1],
+            "a": popt[0] * yn[1] / xn[1],
+            "b": popt[1] * yn[1],
+        }
+
+        return out
+
+    @staticmethod
+    def ramsey(x, y):
```
TheoLaudatQM, line 77: I modified the program quite a lot to make it more robust, readable and with uncertainty in the fitted parameters:
`def ramsey(x, y, verbose=0):
    """
      Create a fit to Ramsey experiment of the form

      .. math::
      f(x) = final_offset * (1 - np.exp(-x * (1/T2))) + amp / 2 * (
          np.exp(-x * (1/tau))
          * (initial_offset * 2 + np.cos(2 * np.pi * f * x + phase))
          )

      for unknown parameters :
          f - The detuning frequency [GHz]
          phase - The phase [rad]
          T2 - The decay constant [ns]
          amp - The amplitude [a.u.]
          final_offset -  The offset visible for long dephasing times [a.u.]
          initial_offset - The offset visible for short dephasing times

      :param x: The dephasing time [ns]
      :param y: Data containing the Ramsey signal
      :param verbose: if 1 prints data into the terminal
      :return: A dictionary of (fit_func, f, phase, tau, amp, uncertainty_population, initial_offset)

       """

    # Normalizing the vectors
    xn = preprocessing.normalize([x], return_norm=True)
    yn = preprocessing.normalize([y], return_norm=True)
    x = xn[0][0]
    y = yn[0][0]

    # Compute the FFT for guessing the frequency
    fft = np.fft.fft(y)
    f = np.fft.fftfreq(len(x))
    # Take the positive part only
    fft = fft[1: len(f)//2]
    f = f[1: len(f)//2]
    # Remove the DC peak if there is one
    if (np.abs(fft)[1:] - np.abs(fft)[:-1] > 0).any():
        first_read_data_ind = np.where(np.abs(fft)[1:] - np.abs(fft)[:-1] > 0)[0][0]  # away from the DC peak
        fft = fft[first_read_data_ind:]
        f = f[first_read_data_ind:]

    # Finding a guess for the frequency
    out_freq = f[np.argmax(np.abs(fft))]
    guess_freq = out_freq / (x[1] - x[0])

    # The period is 1 / guess_freq --> number of oscillations --> peaks decay to get guess_T2
    period = int(np.ceil(1 / out_freq))
    peaks = np.array([np.std(y[i * period: (i + 1) * period]) for i in range(round(len(y) / period))]) * np.sqrt(2) * 2

    # Finding a guess for the decay (slope of log(peaks))
    if len(peaks) > 1:
        guess_T2 = -1 / ((np.log(peaks)[-1] - np.log(peaks)[0]) / (period * (len(peaks) - 1))) * (x[1] - x[0])
    else:
        guess_T2 = 100 / xn[1][0]
        print(Warning("WARNING: The initial guess for the decay failed, increasing the number of oscillations should solve the issue."))

    # Finding a guess for the offsets
    initial_offset = np.mean(y[:period])
    post_decay_mean = np.mean(y[-period:])

    # Finding a guess for the phase
    guess_phase = np.angle(fft[np.argmax(np.abs(fft))]) - guess_freq * 2 * np.pi * x[0]

    # Fitting function
    fit_type = lambda x, a: post_decay_mean * a[4] * (1 - np.exp(-x / guess_T2 * a[1])) + peaks[0] / 2 * a[2] * (
            np.exp(-x / guess_T2 * a[1])
            * (a[5] * initial_offset / peaks[0] * 2 + np.cos(2 * np.pi * a[0] * guess_freq * x + a[3]))
    )
    if verbose:
        print(
            f"Initial guess:\n f = {guess_freq / xn[1][0]}, phase = {guess_phase}, T2 = {guess_T2 * xn[1][0]}, amp = {peaks[0] * yn[1][0]}, uncertainty population = {post_decay_mean * yn[1][0]}, initial offset = {initial_offset * yn[1][0]}"
        )
    import scipy
    def func(x, a0, a1, a2, a3, a4, a5):
        return post_decay_mean * a4 * (1 - np.exp(-x / guess_T2 * a1)) + peaks[0] / 2 * a2 * (
                np.exp(-x / guess_T2 * a1)
                * (a5 * initial_offset / peaks[0] * 2 + np.cos(2 * np.pi * a0 * guess_freq * x + a3))
        )

    popt, pcov = scipy.optimize.curve_fit(
        func,
        x,
        y,
        p0=[1, 1, 1, guess_phase, 1, 1],
    )

    perr = np.sqrt(np.diag(pcov))

    out = {
        "fit_func": lambda x: fit_type(x/xn[1][0], popt) * yn[1][0],
        "f": popt[0] * guess_freq / xn[1][0],
        "phase": popt[3] % (2 * np.pi),
        "T2": (guess_T2 * popt[1]) * xn[1][0],
        "amp": peaks[0] * popt[2] * yn[1][0],
        "uncertainty_population": post_decay_mean * popt[4] * yn[1][0],
        "initial_offset": popt[5] * initial_offset * yn[1][0],
    }
    if verbose:
        print(
            f"Fit results:\n"
            f"f = {out['f']*1000:.3f} +/- {perr[0] * guess_freq / xn[1][0]:.3f} MHz, \n"
            f"phase = {out['phase']:.3f} +/- {perr[3] % (2 * np.pi):.3f} rad, \n"
            f"T2 = {out['T2']:.2f} +/- {perr[1] * guess_T2 * xn[1][0]:.3f} ns, \n"
            f"amp = {out['amp']:.2f} +/- {perr[2] * peaks[0] * yn[1][0]:.3f} a.u., \n"
            f"uncertainty population = {out['uncertainty_population']:.2f} +/- {perr[4] * post_decay_mean * yn[1][0]:.3f} a.u., \n"
            f"initial offset = {out['initial_offset']:.2f} +/- {perr[5] * initial_offset * yn[1][0]:.3f}"
        )
    return out`

We can discuss it whenever you want

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -31,154 +32,321 @@ def linear(x, y):
...
+            plt.plot(x_data, fit_type(x, popt) * y_normal)
+            plt.plot(x_data, y_data, ".", label=f"T2  = {out['T2'][0]:.1f} +/- {out['T2'][1]:.1f}ns")
+            plt.xlabel("Waiting time [ns]")
+            plt.ylabel("I & Q amplitude [a.u.]")
+            plt.legend(loc="upper right")
+
+        if save:
+            fit_params = dict(itertools.islice(out.items(), 1, len(out)))
+            for key in fit_params:
+                fit_params[key] = fit_params[key]
```
TheoLaudatQM, line 444: Why do you need this? I don't see what it does

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -350,22 +358,23 @@ def ramsey(x_data, y_data, guess, verbose=False, plot=False, save=False):
         if guess is not None:
             for key in guess.keys():
                 if key == "f":
-                    guess_freq = guess[key] * x_normal
+                    guess_freq = float(guess[key]) * x_normal
```
TheoLaudatQM, line 361: Why do we need to cast into float here ?

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -511,22 +523,22 @@ def transmission_resonator_spectroscopy(
...
+        # Print the initial guess if verbose=True
         if verbose:
             print(
-                f"Initial guess:\n"
-                f" f = {f0 * x_normal}, \n"
-                f" kc = {(peak - v0)/width0 * y_normal / x_normal}, \n"
-                f" k = {width0 * x_normal}, \n"
+                f"Initial guess:\n "
+                f" f = {f0 * x_normal}, \n "
+                f" kc = {(peak - v0) * (width0 * x_normal) * y_normal}, \n "
```
TheoLaudatQM, line 540: Shouldn't it be `{(peak - v0) / (width0 * x_normal) * y_normal}` instead?

qualang_tools/plot/Fitting/fitting.py
```diff
@@ -663,26 +671,27 @@ def reflection_resonator_spectroscopy(
...
             print(
-                f"Initial guess:\n"
-                f" f = {f0 * x_normal}, \n"
-                f" kc = {(peak - v0) / width0 * y_normal / x_normal}, \n"
-                f" k = {width0 * x_normal}, \n"
-                f" offset = {v0 * y_normal}, \n"
-                f" slope = {m * y_normal / x_normal}"
+                f"Initial guess:\n "
+                f" f = {f0 * x_normal}, \n "
+                f" kc = {(v0 - peak) * (width0 * x_normal) * y_normal}, \n "
```
TheoLaudatQM, line 691: same here `{(peak - v0) / (width0 * x_normal) * y_normal}` ?

qualang_tools/plot/Fitting/fitting.py
MichalGQM: You're right .. I will remove this

qualang_tools/plot/Fitting/fitting.py
MichalGQM: Since when I tried it I wore something like 
guess = {'a': '5'
               'b':'10}
So in case that someone will write it like this ...

qualang_tools/plot/Fitting/fitting.py
MichalGQM: since it is kc/k in order to get kc I need to do kc/k*k

qualang_tools/plot/Fitting/fitting.py
MichalGQM: same... you can check it and let me know if I  am mistaken ...

qualang_tools/plot/README.md
```diff
@@ -21,3 +21,68 @@ while job.result_handles.is_processing():
...
+* Fit the data to a suitable fitting function.
+* Plot the data as well as the fitting function.
+* Extract physical parameters from the fitting.
+* Save the fitting parameters to a json file.
+* Open the json file and save its data to a python dictionary.
+
+The current available fitting functions in the class are:
+
+* Linear fitting.
+* Fitting for the Ramsey experiment (when scanning the dephasing time).
```
TheoLaudatQM, line 37: I think we need to add T1 as well here

qualang_tools/plot/Fitting/fitting.py
TheoLaudatQM: Yes but in python it doesn't matter no? `int * float` will always return a `float`

qualang_tools/plot/Fitting/fitting.py
TheoLaudatQM: Ok I believe you, it is just that it was the inverse before the commit this is why I was confused

qualang_tools/plot/Fitting/fitting.py
MichalGQM: But if it is a string, float(guess[key]) will change it to float...

qualang_tools/plot/Fitting/fitting.py
MichalGQM: Yeah I know.. I checked it but double check is always good :)

qualang_tools/plot/README.md
```diff
@@ -55,38 +55,30 @@ In the example below you can see five parts:
...
+            np.exp(-x * (1/T2))
+            * (initial_offset * 2 + np.cos(2 * np.pi * f * x + phase))
+            )
+
+
+tau_min = 4
+tau_max = 100
+dtau = 2
+taus = np.arange(tau_min, tau_max + 0.1, dtau)
+amplitude = ramsey_data(8 * taus, final_offset=0.1, T2=300, amp=1, initial_offset=0, f=10e-3, phase=0.01) + np.random.normal(
```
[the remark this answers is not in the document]
TheoLaudatQM: Yes I can explain :)
Since I took the data in QUA, taus was in clock cycles so I multiply it by 4 and for the last plot it was actually an echo so I added this additional factor of 2 (8 in total) to convert taus in ns.
Here I am assuming that tau is already in ns so no need for this extra conversion

qualang_tools/plot/Fitting/fitting.py
TheoLaudatQM: Oh right, this is a good point !

qualang_tools/plot/Fitting/fitting.py
TheoLaudatQM: To be sure we need to implement unit tests, do you know how to do it?

MichalGQM: no.. can I see it somewhere or do you want to hop on a call?

qualang_tools/plot/README.md
MichalGQM: You're right

qualang_tools/plot/README.md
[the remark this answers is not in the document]
TheoLaudatQM: because in the way we define it you play pi/2 - tau - pi - tau - pi/2 so the total time is 2*tau

qualang_tools/plot/README.md
```diff
@@ -64,11 +64,11 @@ def ramsey_data(x, final_offset, T2, amp, initial_offset, f, phase):
...
 
-tau_min = 4
-tau_max = 100
-dtau = 2
+tau_min = 32
+tau_max = 800
+dtau = 16
 taus = np.arange(tau_min, tau_max + 0.1, dtau)
-amplitude = ramsey_data(8 * taus, final_offset=0.1, T2=300, amp=1, initial_offset=0, f=10e-3, phase=0.01) + np.random.normal(
+amplitude = ramsey_data(taus, final_offset=0.1, T2=300, amp=1, initial_offset=0, f=10e-3, phase=0.01) + np.random.normal(
```
TheoLaudatQM, line 71: Are these values consistent with the plot? 
I am asking because on the figure it is displayed T2= 37.5ns and f=80MHz...

MichalGQM: You're right, I forgot to change it and I did a small change in the example and changed the plot

TheoLaudatQM: Nice !

tests/test_fittinig_class.py
```diff
@@ -71,4 +71,22 @@ def test_reflection_resonator_spectroscopy(kc, k, f, offset, slope):
...
+@pytest.mark.parametrize("phase", [0, 0.01])
+def test_rabi(offset, T, amp, f, phase):
+    fit = Fit()
+    tau_min = 32
+    tau_max = 800
+    d_tau = 16
+    x = np.arange(tau_min, tau_max + 0.1, d_tau)
+    y = amp * (np.sin(0.5 * (2 * np.pi * f) * x + phase))**2 * np.exp(-x / T) + offset
+
+    out = fit.ramsey(x_data=x, y_data=y)
```
TheoLaudatQM, line 90: This should be fit.rabi I guess

qualang_tools/plot/fitting.py
```diff
@@ -811,6 +811,193 @@ def fit_type(x_var, a):
...
+            for key in guess.keys():
+                if key == "f":
+                    guess_freq = float(guess[key]) * x_normal
+                elif key == "phase":
+                    guess_phase = float(guess[key])
+                elif key == "T2":
+                    guess_T = float(guess[key]) * x_normal
+                elif key == "amp":
+                    peaks[0] = float(guess[key]) / y_normal
+                elif key == "final_offset":
```
TheoLaudatQM, line 921: These parameters don't match the ones from the docstring

TheoLaudatQM (changes requested):
Almost there :)

TheoLaudatQM (approved):
Unless I missed something, I think it is ready to be merged!
Do you agree @yomach ?

PR #132, opened 2022-08-03
allow qm-qua 0.4.0

yomach:


TheoLaudatQM (approved):
Out of curiosity, what is new in 0.4.0?

yomach:
Most the octave API (this is why we have a new major version).
This is the (not final) changelog:
## [Unreleased]
### Fixed
- QuantumMachineManager - Fixed a bug where you could not connect using SSL on python version 3.10+
### Added
- API to control Octave - an up-conversion and down-conversion module with built-in Local Oscillator (LO) sources.
- Support Numpy as input - Support numpy scalars and arrays as valid input. Numpy object can now be used interchanably with python scalars and lists. This apply to all statements imported with `from qm.qua import *`

Can you please release a new py-qua-tools version? We can wait a day or two if Satya is about to merge some stuff.

PR #135, opened 2022-08-26
small fix on waveform_tools for type inference

KosukeMizuno:
The default arguments for `drag_*_pulse_waveforms` were as the following:
```python
def drag_cosine_pulse_waveforms(
    amplitude, length, alpha, anharmonicity, detuning=0, **kwargs
):
...
```

Type inference (for example by pylance) analyzes it as `detuning: int`. Thus, an example usage of `drag_gaussian_pulse_waveforms(..., detuning=1.5e6, ...)` has been marked as wrong type was passed. I just changed `detuing=0` to `detuning=0.0`.

### DIFF
--- a/qualang_tools/config/waveform_tools.py
+++ b/qualang_tools/config/waveform_tools.py
@@ -8,7 +8,7 @@ def drag_gaussian_pulse_waveforms(
     sigma,
     alpha,
     anharmonicity,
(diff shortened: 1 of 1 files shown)


yomach (approved):
Perfect, thanks!

PR #140, opened 2022-09-21
multi user tools: qm session

liorella-qm:
added multi user execution tools, imported from qw-qm lib.

IMPORTANT: Please check that it actually executes because I don't have access currently to a real OPX

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
-- multi user tools subpackage, with a qm_session method that allows several users to work simultaneously.
+- Multi-user Tools - A subpackage that allows several users to work simultaneously.
 
 ## [0.12.0] - 2022-08-29
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
+- multi user tools subpackage, with a qm_session method that allows several users to work simultaneously.
 
 ## [0.12.0] - 2022-08-29
 ### Changed
--- a/qualang_tools/multi_user/README.md
+++ b/qualang_tools/multi_user/README.md
@@ -1,0 +1,30 @@
+# Multi user tools
+
+This subpackage contains tools to be used when multiple users need to work
+on the same QOP, and the queue functionality is insufficient since these
+users need to work with different QMs or configs that share resources.
+
+Currently it contains a single method, `qm_session`, that allows a user to _try_ to
+open a quantum machine, and if it is not possible since its resources are currently in use,
+wait for them to free up. This is done by repeatedly polling the QM manager.
+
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
+- multi user tools subpackage, with a qm_session method that allows several users to work simultaneously.
```
yomach, line 10:
```suggestion
- Multi-user Tools - A subpackage that allows several users to work simultaneously.
```

WDYT?

qualang_tools/multi_user/README.md
```diff
@@ -0,0 +1,30 @@
+# Multi-user tools
```
yomach, line 1: Nice,
We also need to add it to the main readme file

liorella-qm: added

yomach (commented):
Looks good, passing the Stick to Theo

yomach:
@TheoLaudatQM any comments before we merge?

TheoLaudatQM:
I was waiting to figure out the issue when opening two QMs with the same config that we discussed in order to write the unit tests, but we can merge now if it is urgent and write the tests later

yomach:
The behavior is known, and even if we'll change/fix it, it won't be for at least 1-2 months.
Write the tests for the current situation.

TheoLaudatQM:
Ok, I didn't know the status...
I'll write them today so we can merge

yomach:
Changes to gateway will always take time...

yomach:
Thanks!

TheoLaudatQM:
I just modified the file so that the QM is not closed if a job is running while reaching the "finally:" section.
If it is fine for you we can merge

yomach (approved):

TheoLaudatQM (approved):

liorella-qm:
@TheoLaudatQM what is the use case you have in mind for keeping the QM open if a job is running when reaching the `finally` section?

TheoLaudatQM:
From Slack: 
Theo:
>  I was a bit confused to see that when you don't use the stream processing, you directly reach the finally statement which closes the QM even if a job is running. I am not sure whether closing the QM would interrupt the running job or not (but I think it does), but if someone else tries to open the QM then he will be able to do it which would interrupt the running job and I don't think this is an intended behavior

Lior:
>  I see. In my code I do everything related to the job inside the code block of the with statement. If the example doesn't it do it it's wrong… the idea is that you do everything you need the QM for inside this block, and then free it. Then it makes it clearer to the user that they're only getting the QM for a limited time. That’s the basic idea

Theo:
> I see, so we should only use it when we fetch data from the OPX within the context manager. For AMO for instance, since the OPX doesn't measure we usually don't need to use the stream processing even though the OPX is constantly sending pulses

Lior:
> Actually, what happens when you exit the with block with what you added? It'll just stay suspended until the job is finished no?

Theo:
> Yes exactly it will remain in the context manager until the job is done

Lior:
> Yeah so I think you're right, it’s not going to change my use case and if it's a better fit for AMO, let’s do it

PR #141, opened 2022-09-30
Improve the plotting tools

TheoLaudatQM:
Add 1D and 2D plots of I&Q
Add simulated samples plots
Add plot AR attempts

### DIFF
--- a/qualang_tools/plot/README.md
+++ b/qualang_tools/plot/README.md
@@ -93,12 +93,3 @@ job = qmm.simulate(config, test_prog, simulation_config)
 fig = plot_simulator_output([["RF"],["qubit"]], job, config, duration_ns=8000)
 fig.show()
 ```
-
-## plot_ar_attempts
-This function plots the histogram of the number of attempts necessary to perform active reset.
-
-### Usage example
-
-```python
-
-```
--- a/qualang_tools/plot/README.md
+++ b/qualang_tools/plot/README.md
@@ -48,7 +48,7 @@ plot_demodulated_data_2d(time, amp, I, Q, "time [ns]", "amplitude [mV]", "2D map
 ```
 
 ## get_simulated_samples_by_element
-This function gets the samples generated from the QUA simulator element per element.
+This function gets the samples generated from the QUA simulator by element.
 
 ### Usage example
 
@@ -72,7 +72,7 @@ resonator_samples = get_simulated_samples_by_element("resonator", job, config)
 
 
 ## plot_simulator_output
-This function plots the samples generated from the QUA simulator element per element using *plotly*.
+This function plots the samples generated from the QUA simulator by element using *plotly*.
 
 ### Usage example
 
--- a/qualang_tools/plot/README.md
+++ b/qualang_tools/plot/README.md
@@ -1,5 +1,5 @@
 # Plot tools
-This library includes tools to help handling plots from QUA programs.
+This library includes tools to help to handle plots from QUA programs.
 
 ## interrupt_on_close
 This function allows to interrupt the execution and free the console when closing the live-plotting figure.
--- a/qualang_tools/plot/README.md
+++ b/qualang_tools/plot/README.md
@@ -7,17 +7,98 @@ This function allows to interrupt the execution and free the console when closin
 ### Usage example
 
 ```python
+from qm.QuantumMachinesManager import QuantumMachinesManager
 from qualang_tools.plot import interrupt_on_close
+from qm.qua import *
+import matplotlib.pyplot as plt
+from configuration import config
+
+with program() as test_prog:
+    # The QUA program to execute
+    ...
 
 qmm = QuantumMachinesManager()
 qm = qmm.open_qm(config)
-job = qm.execute(cryoscope)
+job = qm.execute(test_prog)
 
-fig = plt.figure(figsize=(15, 15))
+fig = plt.figure()
 interrupt_on_close(fig, job)
 
 while job.result_handles.is_processing():
     # Live plotting
     ...
     # The execution will be cleanly interrupted by closing the figure
+```
+
+## plot_demodulated_data 1D and 2D
+These functions plot the demodulated data (either 'I' and 'Q' or amplitude and phase) from a 1D or 2D scan.
+
+### Usage example
+
+```python
+from qualang_tools.plot import plot_demodulated_data_1d, plot_demodulated_data_2d
+import numpy as np
+
+time = np.linspace(0,1000, 1001)
+amp = np.linspace(-2, 2, 101)
+I = np.array([[np.sin(i/100)*np.cos(k) for i in time] for k in amp])
+Q = np.array([[np.cos(i/100)*np.sin(k) for i in time] for k in amp])
+
+plot_demodulated_data_2d(time, amp, I, Q, "time [ns]", "amplitude [mV]", "2D map", amp_and_phase=True, plot_options={"cmap": "magma"})
+```
+
+## get_simulated_samples_by_element
+This function gets the samples generated from the QUA simulator element per element.
+
+### Usage example
+
+```python
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from qm import SimulationConfig
+from qm.qua import *
+from configuration import config
+from qualang_tools.plot import get_simulated_samples_by_element
+
+with program() as test_prog:
+    # The QUA program to simulate
+    ...
+
+qmm = QuantumMachinesManager()
+simulation_config = SimulationConfig(duration=8000)
+job = qmm.simulate(config, test_prog, simulation_config)
+qubit_samples = get_simulated_samples_by_element("qubit", job, config)
+resonator_samples = get_simulated_samples_by_element("resonator", job, config)
+```
+
+
+## plot_simulator_output
+This function plots the samples generated from the QUA simulator element per element using *plotly*.
+
+### Usage example
+
+```python
+from qm.QuantumMachinesManager import QuantumMachinesManager
+from qm import SimulationConfig
+from qm.qua import *
+from configuration import config
+from qualang_tools.plot import plot_simulator_output
+
+with program() as test_prog:
+    # The QUA program to simulate
+    ...
+
+qmm = QuantumMachinesManager()
+simulation_config = SimulationConfig(duration=8000)
+job = qmm.simulate(config, test_prog, simulation_config)
+fig = plot_simulator_output([["RF"],["qubit"]], job, config, duration_ns=8000)
+fig.show()
+```
+
+## plot_ar_attempts
+This function plots the histogram of the number of attempts necessary to perform active reset.
+
+### Usage example
+
+```python
+
 ```
--- a/qualang_tools/plot/plot.py
+++ b/qualang_tools/plot/plot.py
@@ -1,4 +1,4 @@
-"""Tools to help handling plots from QUA programs.
+"""Tools to help to handle plots from QUA programs.
 
 Content:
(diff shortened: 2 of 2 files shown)


qualang_tools/plot/README.md
```diff
@@ -1,23 +1,104 @@
...
+
+qmm = QuantumMachinesManager()
+simulation_config = SimulationConfig(duration=8000)
+job = qmm.simulate(config, test_prog, simulation_config)
+fig = plot_simulator_output([["RF"],["qubit"]], job, config, duration_ns=8000)
+fig.show()
+```
+
+## plot_ar_attempts
+This function plots the histogram of the number of attempts necessary to perform active reset.
```
yomach, line 98: missing?

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+
+    :param x: 1D numpy array representing the x-axis.
+    :param y: 1D numpy array representing the y-axis.
+    :param I: 2D numpy array representing the 'I' quadrature.
+    :param Q: 2D numpy array representing the 'Q' quadrature.
+    :param x_label: name for the x-axis label.
+    :param y_label: name for the y-axis label.
+    :param title: title of the plot.
+    :param amp_and_phase: if True, plots the amplitude [np.sqrt(I**2 + Q**2)] and phase [signal.detrend(np.unwrap(np.angle(I + 1j * Q)))] instead of I and Q.
+    :param fig: the matplotlib figure if defined prior to calling the function.
```
yomach, line 39:
```suggestion
    :param fig: a matplotlib figure. If `none` (default), will create a new one.
```

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+    :param plot_options: dictionary containing various plotting options ['fontsize', 'cmap', 'figsize'].
```
yomach, line 40:
```suggestion
    :param plot_options: dictionary containing various plotting options. Defaults are ['fontsize': 14, 'cmap': `magma`, 'figsize': None].
```

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+
+    :param x: 1D numpy array representing the x-axis.
+    :param I: 1D numpy array representing the 'I' quadrature.
+    :param Q: 1D numpy array representing the 'Q' quadrature.
+    :param x_label: name for the x-axis label.
+    :param title: title of the plot.
+    :param amp_and_phase: if True, plots the amplitude [np.sqrt(I**2 + Q**2)] and phase [signal.detrend(np.unwrap(np.angle(I + 1j * Q)))] instead of I and Q.
+    :param fig: the matplotlib figure if defined prior to calling the function.
+    :param plot_options: dictionary containing various plotting options ['fontsize', 'color', 'marker', 'linewidth', 'figsize'].
+    :return: the matplotlib figure object.
```
yomach, line 112: same as above

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+    Plot 2D maps of either 'I' and 'Q' or the corresponding amplitude and phase in a single figure.
+
+    :param x: 1D numpy array representing the x-axis.
+    :param y: 1D numpy array representing the y-axis.
+    :param I: 2D numpy array representing the 'I' quadrature.
+    :param Q: 2D numpy array representing the 'Q' quadrature.
+    :param x_label: name for the x-axis label.
+    :param y_label: name for the y-axis label.
+    :param title: title of the plot.
+    :param amp_and_phase: if True, plots the amplitude [np.sqrt(I**2 + Q**2)] and phase [signal.detrend(np.unwrap(np.angle(I + 1j * Q)))] instead of I and Q.
```
yomach, line 38:
```suggestion
    :param amp_and_phase: if `True` (default value), plots the amplitude [np.sqrt(I**2 + Q**2)] and phase [signal.detrend(np.unwrap(np.angle(I + 1j * Q)))] instead of I and Q.
```

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+    I: np.ndarray,
+    Q: np.ndarray,
+    x_label: str = None,
+    title: str = None,
+    amp_and_phase: bool = True,
+    fig=None,
+    plot_options: dict = None,
+):
+    """
+    Plot 1D graphs of either 'I' and 'Q' or the corresponding amplitude and phase in a single figure.
```
yomach, line 102:
```suggestion
    Plots 1D graphs of either 'I' and 'Q' or the corresponding amplitude and phase in a single figure.
```

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+    )
+    plt.xlabel(x_label, fontsize=_plot_options["fontsize"])
+    plt.ylabel(y2_label, fontsize=_plot_options["fontsize"])
+    plt.xticks(fontsize=_plot_options["fontsize"])
+    plt.yticks(fontsize=_plot_options["fontsize"])
+    plt.tight_layout()
+    return fig
+
+
+def plot_channel(ax, data: np.ndarray, title: str):
```
yomach, line 173: I don't get this function, this is the same as just using normal plot, but it decides on the `ylim`?

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+def get_simulated_samples_by_element(element_name: str, job: QmJob, config: dict):
+    """
+    Extract the simulated samples corresponding to a given element.
+
+    :param element_name: name of the element corresponding to the samples to extract. Must be defined in the config.
+    :param job: The simulated QmJob from which the samples will be extracted.
+    :param config: The config file used to create the job.
+    :return: 1D complex array containing the simulated samples for the specified element.
+    """
+    element = config["elements"][element_name]
```
yomach, line 201: Probably need to add here a `wait_for_all_values()`

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+        port = element["singleInput"]["port"]
+        samples = sample_struct.__dict__[port[0]].analog[str(port[1])]
+    return samples
+
+
+def plot_simulator_output(
+    plot_axes: List[List[str]], job: QmJob, config: dict, duration_ns: int
+):
+    """
+    generate a 'plotly' plot of simulator output by elements
```
yomach, line 220:
```suggestion
    Generates a 'plotly' plot of simulator output by elements
```

qualang_tools/plot/plot.py
```diff
@@ -1,17 +1,296 @@
...
+                        x=time_vec, y=samples_struct[i][j].imag, name=plot_item + " Q"
+                    ),
+                    row=i + 1,
+                    col=1,
+                )
+    fig.update_xaxes(title="time [ns]")
+    return fig
+
+
+def plot_ar_attempts(res_handles, qubits: List[int], attempts_data: str, **hist_kwargs):
```
yomach, line 266: This is a very specific function, not sure what I think about it...
Does it matches our active reset code? Did we implement the repeat-until-success protocol in a manner which is compatible with this?
If so, we should link between the functions, this is the syntax
```
:func:`~qm.qua._dsl.declare`
```

qualang_tools/plot/README.md
TheoLaudatQM: After discussing it with Lior, I will not add this tools since it is very specific and customers most likely don't need it

qualang_tools/plot/plot.py
TheoLaudatQM: thanks

qualang_tools/plot/plot.py
TheoLaudatQM: thanks

qualang_tools/plot/plot.py
TheoLaudatQM: thanks

qualang_tools/plot/plot.py
TheoLaudatQM: I never know about this one, shall we always use the 3rd person or the infinitive?

qualang_tools/plot/plot.py
TheoLaudatQM: Yes, it was in Lior's lib but I won't add it because it is not very useful and we already have plot 1D and 2D now

qualang_tools/plot/plot.py
TheoLaudatQM: All what happens here is meant to execute once the simulation is done, so I would prefer to put the `wait_for_all_values()` in the program rather than in the function 
Also, I am not sure that we need it since we just want to extract the simulated samples and not something from `stream_processing`, what do you think?

qualang_tools/plot/plot.py
TheoLaudatQM: No we currently don't and I'll remove this function which was not asked by customers (it will remain in Lior's lib though)

qualang_tools/plot/plot.py
yomach: Just needs to be consistent. Not sure which is better

qualang_tools/plot/plot.py
yomach: The information from the simulation will also not be available until the `wait_for_all_values()` is finished.
So if you write in a code:
```python
job = qmm.simulate(...)
get_simulated_samples_by_element(...)
```
You'll get an error...

TheoLaudatQM: I don't see why we need `wait_for_all_values()` here...
Maybe it depends on the versions, but I retested it and I don't have any error without it

yomach: Did you simulate a program with stream processing? And let it run to actually get a bit of data?
You don't have access to the simulator samples until the SP is done. Trust me...
If your program didn't had it, or very simple SP, then it might work without (due to the python delay just being long enough). But in many programs it will give an error...

TheoLaudatQM: Yes, but it was fairly simple (resonator spec) so I couldn't get any error, no matter the simulation duration or number of points...
But I believe you and I will add the `wait_for_all_values()` which cannot harm anyway

qualang_tools/plot/README.md
```diff
@@ -1,23 +1,95 @@
...
+time = np.linspace(0,1000, 1001)
+amp = np.linspace(-2, 2, 101)
+I = np.array([[np.sin(i/100)*np.cos(k) for i in time] for k in amp])
+Q = np.array([[np.cos(i/100)*np.sin(k) for i in time] for k in amp])
+
+plot_demodulated_data_2d(time, amp, I, Q, "time [ns]", "amplitude [mV]", "2D map", amp_and_phase=True, plot_options={"cmap": "magma"})
+```
+
+## get_simulated_samples_by_element
+This function gets the samples generated from the QUA simulator element per element.
```
yomach, line 51: element per element is not clear.
Maybe "by element"?

qualang_tools/plot/README.md
```diff
@@ -1,23 +1,95 @@
...
+qmm = QuantumMachinesManager()
+simulation_config = SimulationConfig(duration=8000)
+job = qmm.simulate(config, test_prog, simulation_config)
+qubit_samples = get_simulated_samples_by_element("qubit", job, config)
+resonator_samples = get_simulated_samples_by_element("resonator", job, config)
+```
+
+
+## plot_simulator_output
+This function plots the samples generated from the QUA simulator element per element using *plotly*.
```
yomach, line 75: same

yomach (approved):

PR #142, opened 2022-10-10
Config helper tools

TheoLaudatQM:


qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+        #         self.data["waveforms"][wf_name + "_i"] = {
+        #             "type": "arbitrary",
+        #             "samples": list(wf_i),
+        #         }
+        #         self.data["waveforms"][wf_name + "_q"] = {
+        #             "type": "arbitrary",
+        #             "samples": list(wf_q),
+        #         }
+        # else:
+        #     raise ValueError("The waveforms must have at least one point.")
```
yomach, line 83: removed commented lines

qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+        #             "type": "constant",
+        #             "sample": wf[0],
+        #         }
+        #     else:
+        #         self.data["waveforms"][wf_name] = {
+        #             "type": "arbitrary",
+        #             "samples": list(wf),
+        #         }
+        # else:
+        #     raise ValueError("The waveform must have at least one point.")
```
yomach, line 126: same

qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+        if operation_name not in self.data["elements"][element]["operations"].keys():
+            raise KeyError(
+                f"The operation '{operation_name}' in not defined in the config for the element '{element}'."
+            )
+
+        return self.data["pulses"][
+            self.data["elements"][element]["operations"][operation_name]
+        ]
+
+    def update_op_amp(self, element: str, operation_name: str, amp: float):
```
yomach, line 191: This changes the waveform amplitude, which can affect other elements as well!
This has to be at least noted... Options:
1. Just note it.
2. Look to see if anyone else is using that waveform (search through the pulses/operations) and if so, give a warning/error/option 3 from below
3. Copy the waveform and create a new one with the new amplitude

We can also have a flag indicating the behavior (between 1 & 3, or between 2 & 3 if we decide to do it)

qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+
+        pulse = self.get_pulse_from_op(element, operation_name)
+        try:
+            return self.data["waveforms"][pulse["waveforms"]["single"]]["sample"]
+        except KeyError:
+            raise KeyError(
+                "Can only access amplitude for a constant pulse of a single element"
+            )
+
+    def update_integration_weight(
```
yomach, line 246: same as for amplitude, although integration weights are less commonly shared

tests/test_config_helper_tools.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+            wait(cooldown_time, "resonator")
+            save(I, I_st)
+            save(Q, Q_st)
+
+    with stream_processing():
+        I_st.buffer(len(freqs)).average().save("I")
+        Q_st.buffer(len(freqs)).average().save("Q")
+
+#####################################
```
yomach, line 173: removed commented lines

qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+
+        pulse_name = self.data["elements"][element]["operations"][operation_name]
+        self.data["pulses"][pulse_name]["length"] = len(wf[0])
+        # mixInputs element
+        if len(wf) == 2:
+            wf_name = element + "_" + operation_name + self.wf_suffix
+            wf_i = wf[0]
+            wf_q = wf[1]
+            if len(wf_i) > 0:
+                if wf_i[:-1] == wf_i[1:]:
```
yomach, line 346: I think that you are checking if the waveforms are symmetric, not if they are constant. For example, a gaussian or a cosine wave would be detected here as "constant".
This has some nice solutions for checking if all elements are equal:
https://stackoverflow.com/questions/3844801/check-if-all-elements-in-a-list-are-identical

qualang_tools/config/helper_tools.py
```diff
@@ -0,0 +1,390 @@
...
+                    self.data["waveforms"][wf_name + "_i"] = {
+                        "type": "constant",
+                        "sample": wf_i[0],
+                    }
+                    self.data["waveforms"][wf_name + "_q"] = {
+                        "type": "constant",
+                        "sample": wf_q[0],
+                    }
+                else:
```
yomach, line 355: Also, this has to be checked separately for I & Q, because it is very common for one to be arbitrary and for the other to be constant...

yomach (commented):
Gave a few comments. Haven't played around with it a bit, so just a simple code review and what I could see from here.

qualang_tools/config/helper_tools.py
TheoLaudatQM: This is a good point, thanks!
Now I check if it used elsewhere and I added a flag to force the update (default is False)

qualang_tools/config/helper_tools.py
TheoLaudatQM: Not really, it compares two lists shifted by one item so if all elements are equal, then it will be True and if the list is symmetric then it will be False because all neighbors won't be equal. You can test it with `wf = [0, 1, 0]` where it compares [0,1] with [1,0]

qualang_tools/config/helper_tools.py
TheoLaudatQM: Good point !

qualang_tools/config/helper_tools.py
TheoLaudatQM: Good point as well, I did the same as for update_op_amp

yomach (approved):

qualang_tools/config/helper_tools.py
yomach: My bad

PR #144, opened 2022-11-11
fixes to bakery

yomach:
- Bakery - Added a flag to disable the addition of an `align` to the `run` function.
- Bakery - Now uses the more efficient `frame_rotation_2pi` instead of `frame_rotation`.
- Bakery - Fixed cases in which a `frame_rotation_2pi(0.0)` was added .


TheoLaudatQM (approved):
I didn't actually test it but the writing seems correct

PR #146, opened 2022-11-16
Update plot_simulator_output

TheoLaudatQM:
Plot digital pulse
Take into account elements with digital inputs only
Allow to plot all elements used in the qua program if plot_axes is not provided

### DIFF
--- a/qualang_tools/plot/plot.py
+++ b/qualang_tools/plot/plot.py
@@ -208,12 +208,12 @@ def get_simulated_samples_by_element(element_name: str, job: QmJob, config: dict
     return analog_samples, digital_samples
 
 def plot_simulator_output(
-    plot_axes: List[List[str]], job: QmJob, config: dict, duration_ns: int
+    job: QmJob, config: dict, duration_ns: int, plot_axes: List[List[str]] = (()),
 ):
     """
     Generate a 'plotly' plot of simulator output by elements
     
-    :param plot_axes: a list of lists of elements (ex: [["qubit0"], ["qubit1"]]). Will open
+    :param plot_axes: a list of lists of elements (ex: [["qubit0", "qubit1"], ["resonator0", "resonator1"]]). Will open
     multiple axes, one for each list.
     :param job: The simulated QmJob to plot.
     :param config: The config file used to create the job.
--- a/qualang_tools/plot/plot.py
+++ b/qualang_tools/plot/plot.py
@@ -196,21 +196,23 @@ def get_simulated_samples_by_element(element_name: str, job: QmJob, config: dict
     else:
         analog_samples = None
     # Digital waveforms
+    digital_samples = []
     if "digitalInputs" in element:
-        port = element["digitalInputs"]["port"]
-        digital_samples = sample_struct.__dict__[port[0]].digital[str(port[1])].astype(int)
+
+        for key in element["digitalInputs"].keys():
+            port = element["digitalInputs"][key]["port"]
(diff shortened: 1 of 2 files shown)


qualang_tools/plot/plot.py
```diff
@@ -170,77 +170,146 @@ def plot_demodulated_data_1d(
...
+    plot_axes: List[List[str]] = (()),
+    **kwargs,
 ):
     """
     Generate a 'plotly' plot of simulator output by elements
 
-    :param plot_axes: a list of lists of elements (ex: [["qubit0"], ["qubit1"]]). Will open
-    multiple axes, one for each list.
+    :param plot_axes: (optional) a list of lists of elements (ex: [["qubit0", "qubit1"], ["resonator0", "resonator1"]]). Will open
+    multiple axes, one for each list. If not provided, then the qua_program must be specified as kwargs using qua_program= and all the elements will be plotted in their own axis.
```
yomach, line 226: Very confusing to have two optional, one in kwargs...
Why not add `qua_program` to the variables as well?

TheoLaudatQM: The thing is we don't need the QUA program except when plot_axes isn't provided and I understand that it can be painful for the user to input the qua program every time...
I tried to get the elements used in the program from the job (instead of the program) but I didn't manage...

qualang_tools/plot/plot.py
```diff
@@ -170,77 +170,146 @@ def plot_demodulated_data_1d(
...
+
+    return analog_samples, digital_samples
 
 
 def plot_simulator_output(
-    plot_axes: List[List[str]], job: QmJob, config: dict, duration_ns: int
+    job: QmJob,
+    config: dict,
+    duration_ns: int,
+    plot_axes: List[List[str]] = (()),
```
yomach, line 219:
```suggestion
    plot_axes: List[List[str]] = (()),
    program = None,
```

qualang_tools/plot/plot.py
yomach: Look at the comment I added.
and then here:
```suggestion
    :param plot_axes: (Either this or 'program' have to be provided) a list of lists of elements (ex: [["qubit0", "qubit1"], ["resonator0", "resonator1"]]). Will open multiple axes, one for each list. 
    :param qua_program: (Either this or 'plot_axes' have to be provided) All the elements will be plotted in their own axis.

```

What does it mean that everyone will be in its own axis? different plots?

TheoLaudatQM: It's a request from a customer actually

yomach: seems a bit extreme to me, but ok.

yomach (approved):

yomach:
Didn't test it, just went over the code

qualang_tools/plot/plot.py
TheoLaudatQM: Why extreme?

yomach: Maybe bloat is a better word?

To me, it seems more friendly to have all of the elements on the same axis. It's hard for me to imagine wanting them on seperate axes, it would take too much space and would be hard to compare and check timing.

A much better option, I think, is having them on the same axis but shifted in the Y axis to be one on top of the other, this is how usually timing is displayed.

TheoLaudatQM: I see, I guess it is for readability issue if you have a lot of elements... And note that with plotly comparing different axes is much better than with matplotlib and I kind of like it this way, you should try it.

PR #147, opened 2022-11-16
guis and independent discriminator 

joseph-qm:


yomach (changes requested):
Without looking at it too deeply, there are a few things that I can already say:
1. Need to remove the "joe_testing" and "active_reset_full" folders
2. Need to update main readme file
3. Need to update changelog
4. Need to add info about the control panel GUI (to readme)
5. Need to remove configuration & IP addresses from the control panel GUI
6. Need to run black formatting on everything

You can take a look at this for a bit more info:
https://github.com/qua-platform/py-qua-tools/blob/main/CONTRIBUTING.md#code-standards

PR #151, opened 2022-12-09
py-qua-tools improvement

TheoLaudatQM:
Fitting tools update and bug fixes
Add fit to 1D plotting function
Allow from array to have 1 point
Add iteration number to progress bar

### DIFF
--- a/qualang_tools/plot/plot.py
+++ b/qualang_tools/plot/plot.py
@@ -98,8 +98,8 @@ def plot_demodulated_data_1d(
     title: str = None,
     amp_and_phase: bool = True,
     fig=None,
-    fit=None,
     plot_options: dict = None,
+    fit=None,
 ):
     """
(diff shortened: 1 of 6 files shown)


qualang_tools/plot/plot.py
```diff
@@ -97,6 +98,7 @@ def plot_demodulated_data_1d(
     title: str = None,
     amp_and_phase: bool = True,
     fig=None,
+    fit=None,
```
yomach, line 101: Adding a new variable in the middle is a breaking change; it should at least be noted in the changelog.

yomach (changes requested):
Looks good (in terms of code, haven't tested the functionality).
One comment to fix.

qualang_tools/plot/plot.py
TheoLaudatQM: I put it as the last input so that it doesn't break, thanks!

yomach (approved):

PR #154, opened 2023-02-20
assign_variables_to_element

yomach:


README.md
```diff
@@ -26,6 +26,8 @@ storing them in the usual configuration file. It allows defining waveforms in a
   * [Interactive Plot Library](qualang_tools/addons/README.md) - This package drastically extends the capabilities of matplotlib,
   enables easily editing various parts of the figure, copy-pasting data between figures and into spreadsheets, 
   fitting the data and saving the figures.
+  * [assign_variables_to_element](qualang_tools/addons/variables.py) - Forces the given variables to be used by the given element thread. Useful as a workaround for when the compiler
+  wrongly assigns variables which can causes gaps.
```
TheoLaudatQM, line 30: No 's' at "which can **cause** gaps"

qualang_tools/addons/variables.py
```diff
@@ -0,0 +1,12 @@
+from qm.qua import *
+
+
+def assign_variables_to_element(element, *variables):
+    """
+    Forces the given variables to be used by the given element thread. Useful as a workaround for when the compiler
+    wrongly assigns variables which can causes gaps.
```
TheoLaudatQM, line 7: No 's' at "which can **cause** gaps"

qualang_tools/addons/variables.py
```diff
@@ -0,0 +1,12 @@
+from qm.qua import *
+
+
+def assign_variables_to_element(element, *variables):
```
TheoLaudatQM, line 4: If I understand correctly, the usage would be:
`a = declare(fixed)`
`b = declare(int)`
 `assign_variables_to_element("qubit_0", a, b)`
`...`
`Real time computation with a and b`
And this will just add a 16ns latency at the beginning of the program to force a given element to be assigned to the variable. Is it correct?

TheoLaudatQM, line 4: Also I would add the expected format for "element" and "variables" in the docstring.
Should "variables" be a list (`assign_variables_to_element("qubit_0", [a, b])`) or just add the variables one after the other (`assign_variables_to_element("qubit_0", a, b)`)?

qualang_tools/addons/variables.py
```diff
@@ -0,0 +1,12 @@
...
+    """
+    Forces the given variables to be used by the given element thread. Useful as a workaround for when the compiler
+    wrongly assigns variables which can causes gaps.
+    """
+    _exp = Cast.to_int(variables[0])
+    for variable in variables[1:]:
```
TheoLaudatQM, line 10: Won't this fail if variables is just a single QUA variable?

TheoLaudatQM (changes requested):

qualang_tools/addons/variables.py
yomach: yes, exactly.

yomach: Will do.
Currently, it's one after the other `assign_variables_to_element("qubit_0", a, b)`.

qualang_tools/addons/variables.py
yomach: No, it works :)

TheoLaudatQM (approved):
All good to me :)

PR #156, opened 2023-02-24
qcodes drivers

TheoLaudatQM:
I put it under py-qua-tools so that people can just download the opx and experiment drivers (that they may want to modify to suit their experiments) but I still would like to provide a usage file (hello_qcodes) and I didn't know where to put it...

### DIFF
--- a/examples/Qcodes_drivers/README.md
+++ b/examples/Qcodes_drivers/README.md
@@ -1,0 +1,15 @@
+# QCoDeS drivers usage example
+
+__Package__: https://github.com/qua-platform/py-qua-tools/tree/main/qualang_tools/external_frameworks/qcodes
+
+This package contains drivers to set the OPX as a QCoDeS instrument and synchronize external parameter sweeps with 
+sequences running on the OPX using the QCoDeS do"nd" methods. 
+Since the OPX contains an FPGA, it can perform complex sequences, acquire data with different formats and sweep 
+parameters on his own, which complexifies a bit the drivers compared to other instruments.
+
+In this example folder, you will find two folders containing:
+* [basic-driver](basic-driver): demonstrating how to use the basic QCoDeS driver for the OPX.
+  * [hello_qcodes.py](basic-driver/hello_qcodes.py): the example file showing the basic driver usage.
+* [advanced-driver](advanced-driver): showing how to modify the basic driver QCoDeS driver.
+  * [advanced_driver.py](advanced-driver/advanced_drivers.py): the modified driver.
+  * [hello_qcodes_advanced.py](advanced-driver/hello_qcodes_advanced.py): the example file showing the advanced driver usage.
--- a/examples/Qcodes_drivers/README.md
+++ b/examples/Qcodes_drivers/README.md
@@ -4,8 +4,9 @@ __Package__: https://github.com/qua-platform/py-qua-tools/tree/main/qualang_tool
 
 This package contains drivers to set the OPX as a QCoDeS instrument and synchronize external parameter sweeps with 
 sequences running on the OPX using the QCoDeS do"nd" methods. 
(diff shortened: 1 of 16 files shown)


yomach:
> I put it under py-qua-tools so that people can just download the opx and experiment drivers (that they may want to modify to suit their experiments) but I still would like to provide a usage file (hello_qcodes) and I didn't know where to put it...

The example should be in `qua-libs`. Previous examples used to be under `examples`, now they are under `examples-old` which we are supposed to slowly either delete or move to a new folder...
Maybe we can open a new examples folder, or to put it under tutorials.

We can also decide to add a dedicated folder under the `py-qua-tools` git for examples, but not under `qualang_tools`. The example file should not be part of the package.

TheoLaudatQM:
I think that the examples should be in the same repo otherwise it feels confusing to me, so I will create an example folder under py-qua-tools. However, since the qcodes implementation depends on the users, they will need to have the ability to modify the files themselves, especially for experimental_drivers.py because it is directly related to the QUA programs they want to run...
Maybe the best would be to have opx_driver.py and experimental_drivers.py in qualang-tools and hello_qcodes.py in examples with a detailed explanation on the structure of experimental_drivers.py so that they will be able to modify it and create their own driver based on the sequences they want to implement, what do you think?

yomach:
Wait, if `experimental_drivers.py` needs to be edited by the user, then it doesn't belong to a package...
Always imagine that `qualang_tools` is like numpy. You just install and use it, you don't edit their files.

lets discuss in person

examples/Qcodes_drivers/README.md
```diff
@@ -0,0 +1,15 @@
+# QCoDeS drivers usage example
+
+__Package__: https://github.com/qua-platform/py-qua-tools/tree/main/qualang_tools/external_frameworks/qcodes
+
+This package contains drivers to set the OPX as a QCoDeS instrument and synchronize external parameter sweeps with 
+sequences running on the OPX using the QCoDeS do"nd" methods. 
+Since the OPX contains an FPGA, it can perform complex sequences, acquire data with different formats and sweep 
+parameters on his own, which complexifies a bit the drivers compared to other instruments.
```
yomach, line 8: See comment in main readme

qualang_tools/external_frameworks/qcodes/README.md
```diff
@@ -0,0 +1,165 @@
+# QCoDeS drivers
+
+This subpackage contains drivers to set the OPX as a QCoDeS instrument and synchronize external parameter sweeps with 
+sequences running on the OPX using the QCoDeS do"nd" methods. 
+Since the OPX contains an FPGA, it can perform complex sequences, acquire data with different formats and 
+sweep parameters on his own, which complexifies a bit the drivers compared to other instruments.
```
yomach, line 6:
```suggestion
Since the OPX runs on a unique pulse processor, it can perform complex sequences that QCoDeS do not natively support. For example, acquire data with different formats and sweep parameters on his own.
To more fully support the OPX capabilities, this driver is more complex compared to other QCoDeS instruments' drivers.
```
Something along these lines.
Note that also QBlox and ZI have an FPGA, it's what you do with it :)

yomach (commented):
Very small cosmetic comments, looks very good!!!
Note I didn't go too deeply into the code, just over the structure and the readme files.

qualang_tools/external_frameworks/qcodes/README.md
TheoLaudatQM: Great I did the modification :)

examples/Qcodes_drivers/README.md
TheoLaudatQM: I did the modification :)

yomach (approved):

yomach (approved):

PR #168, opened 2023-09-04
Fix sign of the rotation angle

TheoLaudatQM:


TheoLaudatQM:
Actually no this doesn't work...

yomach:
The original sign was right, what do you see?
@TheoLaudatQM

TheoLaudatQM:
Yeah I redid all the calculations and it appears that it is actually the way the integration weights are rotated in the config that is wrong. This is why I closed this PR, I opened it too quickly....

PR #172, opened 2023-09-26
Small fixes and live plotting

TheoLaudatQM:
Fix a small issue with raw adc traces and add a live-plotting option when averaging on the most outer loop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### Fixed
 - ConfigBuilder - `Element` now correctly accepts default arguments.
+### Added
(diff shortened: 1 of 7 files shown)


TheoLaudatQM:
Now it also keep the qm_id in case close_other_machines is False and the user wants to update the qm
And there is a new method to update the readout length of a given element and operation

yomach (commented):
Looks good, but are we testing this in any way?

TheoLaudatQM:
I am testing yes, but do you mean with unitary tests?

yomach:
Yes, I mean automated tests in the package.
It'll be good to add them, if possible...
If it's too complicated, then it's an open task

TheoLaudatQM:
Yeah I always had it in the back of my mine, but I always had more pressing things to do and it requires a connection to an OPX our cloud simulator...

yomach (approved):

PR #175, opened 2023-11-27
Add feature for waiting longer than current maximum wait time.

deanpoulos:
Adds a macro to perform a QUA wait command for longer than the current maximum wait time, 2^31-1. 

Also tentatively adds a namespace called `macros`. Open to a better convention than this.


qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,23 @@
+from typing import Union
+from qm.qua import declare, wait, for_
+
+MAX_WAIT = 2 ** 31 - 1  # units of clock cycles (4ns)
+
+
+def long_wait(wait_time: Union[float, int], *elements: str):
```
yomach, line 7: Add a note that only Python variables can be used.

And maybe also raise an error if someone tries?

yomach:
Also, add it to the main readme page

qualang_tools/__init__.py
```diff
@@ -7,6 +7,7 @@
     "results",
     "plot",
     "analysis",
+    "macros",
```
TheoLaudatQM, line 10: @yomach, is there anything preventing this from being in the "addons" folder?

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Note: `wait_time` must strictly be a python literal.
+
+    Args:
+        wait_time (Union[float, int]): Desired wait time in units of clock cycles (4ns).
+        *elements (str): Variable length argument list of elements to wait on.
+    """
+    i = declare(int)
+
+    if not isinstance(wait_time, float | int):
```
TheoLaudatQM, line 20: This can through an error in case of numpy.int32, numpy.int64...
Also, why allowing floats here, since the wait time can only be an integer number of clock cycles? 
Maybe casting it to an integer and raising a warning in case a number with non-zero decimal is provided would be safer, what do you think?

qualang_tools/__init__.py
yomach: No, we can put it there.
We need to think on the structure, it seems that currently almost everything is in it's one file :)

qualang_tools/macros/long_wait.py
yomach: Maybe we can compare against the `_Variable` instance from `qm.qua._dsl`.
It's bad practice, but it is being done in other parts (loops) and there's a ticket to expose it properly:
https://quantum-machines.atlassian.net/browse/QUA-243

TheoLaudatQM: Yes I agree now I am only using `isinstance(variable, (_Variable, _Expression))` to be sure :)

qualang_tools/__init__.py
TheoLaudatQM: I definitely agree and I also don't like having one folder for each tool because it makes it hard for customers to find them :)

TheoLaudatQM:
@deanpoulos do you still want to merge this one?

qualang_tools/__init__.py
deanpoulos: Hopefully it's a good place for people to put macros in the future. Happy to change the structure later

qualang_tools/macros/long_wait.py
TheoLaudatQM: Indeed, if it cannot be a QUA variable, it should be fine

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,46 @@
...
+
+MAX_WAIT = 2**31 - 1  # units of clock cycles (4ns)
+
+
+def long_wait(wait_time: Union[float, int], *elements: str, threshold_for_looping: int = u.to_clock_cycles(1e6)):
+    """Perform a QUA wait command for longer than the current maximum wait time, 2^31-1.
+
+    This macro unrolls into QUA for-loop of smaller waits if the maximum wait time is exceeded in `wait_time`.
+
+    Note: `wait_time` must be a python number.
```
TheoLaudatQM, line 15: A Python `int` I guess then

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,46 @@
+from typing import Union
+
+import numpy as np
+from qm.qua import declare, wait, for_
+from qualang_tools.addons.calibration.calibrations import u
+
+MAX_WAIT = 2**31 - 1  # units of clock cycles (4ns)
+
+
+def long_wait(wait_time: Union[float, int], *elements: str, threshold_for_looping: int = u.to_clock_cycles(1e6)):
```
TheoLaudatQM, line 10: It should only be an `int`, no?

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,46 @@
...
+    """Perform a QUA wait command for longer than the current maximum wait time, 2^31-1.
+
+    This macro unrolls into QUA for-loop of smaller waits if the maximum wait time is exceeded in `wait_time`.
+
+    Note: `wait_time` must be a python number.
+
+    Args:
+        wait_time (Union[float, int]): Desired wait time in units of clock cycles (4ns).
```
TheoLaudatQM, line 18: same here

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,46 @@
...
+        *elements (str): Variable length argument list of elements to wait on.
+        threshold_for_looping: Minimum wait time at which `long_wait` uses a QUA for-loop of waits (default 1ms)
```
TheoLaudatQM, line 20: I would add a higher bound here, because the whole point of this function is that there is a maximum wait time

qualang_tools/macros/long_wait.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    i = declare(int)
```
TheoLaudatQM, line 22: Maybe chose a different name, `i` is often used in QUA loops and I am not sure that it will work if the user does:
```
i = declare(int)
with for_(i, 0, i<10, i+1):
    long_wait() 
```
What do you think?

TheoLaudatQM: And if it works, maybe it is worth adding a test in case it changes in later versions

TheoLaudatQM (changes requested):

qualang_tools/macros/long_wait.py
deanpoulos: Done

qualang_tools/macros/long_wait.py
deanpoulos: Yes

qualang_tools/macros/long_wait.py
deanpoulos: yes

qualang_tools/macros/long_wait.py
deanpoulos: So the reason I did this is because constants above 1ms was causing long compile times. I wrote a unittest for a 30s `long_wait` using this value as the threshold and it worked well. When I was using `MAX_WAIT` as the threshold, it took really long to compile. Have you experienced this before?

qualang_tools/macros/long_wait.py
deanpoulos: good idea.. I will change it to something really specific and just leave a comment noting that it shouldn't be something commonly used.

TheoLaudatQM: Yes perfect!

qualang_tools/macros/long_wait.py
TheoLaudatQM: I see, yeah it already happened to me but I never really understood why...
I am also using 1ms as a safe threshold so it is fine :)

TheoLaudatQM (commented):
All good, we just need to update the main readme and make one for the macros explaining the function and showing basic usage

qualang_tools/macros/long_wait.py
deanpoulos: You make a good point though. Let me add it as a comment to explain why it's 1ms and not higher

deanpoulos:
> All good, we just need to update the main readme and make one for the macros explaining the function and showing basic usage

I added documentation to the READMEs, but not the changelog yet. If you're happy with the READMEs, lets do the changelog together and merge.

TheoLaudatQM (approved):
All good, ready to merge!

PR #176, opened 2023-11-28
Filter calculation tools

yomach:
Hi, take a look at this and please review it :)

It's without any tests and without the changelog/readme changes.

It's also mostly without any hardware limitations (besides the highpass), we need to decide if we want to include them or not. And if so, how to deal with changing limitations.


TheoLaudatQM:
I like the idea of finally putting all of these calculations on the Github very much! 💯 
I looked at the highpass_correction and exponential_correction which I know and it looks good. Do you know the maximum high pass time constant that we can correct based on the resolution of the FIR taps? Is it on 17 or 21 bits? 
I am not familiar with the bounce_and_delay_correction and the filters concatenation, but I trust you :)

yomach:
> I like the idea of finally putting all of these calculations on the Github very much! 💯 I looked at the highpass_correction and exponential_correction which I know and it looks good. Do you know the maximum high pass time constant that we can correct based on the resolution of the FIR taps? Is it on 17 or 21 bits? I am not familiar with the bounce_and_delay_correction and the filters concatenation, but I trust you :)

Thx!
It was actually planned for a long time...

I want to remove the HW limit from the main function (maybe add it as another function) because:
1. We're not even sure what it is (see my slack message)
2. This will change between versions...

TheoLaudatQM:
I see, regarding the HW limitation I think it is now clear that currently the FIR taps are on 25bits and the IIR on 17bits. Maybe we can add flag to discriminate between the QOP (or qm-qua) versions and make these limitations explicit so that the users will know what is going on and if the given taps are accurate or not, what do you think?

TheoLaudatQM:
Hey @yomach, I like this one a lot and many customers are actually asking for such a tool.
What is preventing us from pushing it forward?

yomach:
Not much actually. I guess that we need to:
1. Remove the hardware limitations from `highpass_correction` - make all functions return the ideal ones by default, and then have a separate function that can accept filters and change them according to the hardware limitations (and, it will accept an enum with a version in the future). That's at least the method I had in mind, Open to suggestions*.
2. Add some tests
3. Add to changelog & readme. Go over docstrings (I think I wrote them alright).

Do you have time to do it?

* Maybe we can also have the main function accept an optional qop_version enum that will apply the hardware constrains if given, and return ideal ones if not.

TheoLaudatQM:
Cool!
Yes I can definitely give it a shot :)

CHANGELOG.md
```diff
@@ -12,6 +12,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+- digital_filters - Added a set of functions allowing the derivation the digital filter taps used to correct specific distortions:
+  - ``low_pass_exponential``: Exponential decay defined as `1 + A * exp(-t/tau)`.
+  - ``high_pass_exponential``: Exponential decay defined as `np.exp(-x / t)`.
+  - ``single_exponential_correction``:  Calculate the best FIR and IIR filter taps to correct for an exponential decay (LPF) of the shape
+    `1 + A * exp(-t/tau)`.
+  - ``highpass_correction``: Calculate the best FIR and IIR filter taps to correct for a highpass decay (HPF) of the shape `exp(-t/tau)`.
+  - ``bounce_and_delay_correction``: Calculate the FIR filter taps to correct for reflections (bounce corrections) and to add a delay.
+  - ``calc_filter_taps``: Calculate the best FIR and IIR filter taps for a system with any combination of FIR corrections, exponential
+    corrections (LPF), high pass compensation, reflections (bounce corrections) and a needed delay on the line.
```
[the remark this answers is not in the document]
TheoLaudatQM: good point, I fixed it!

qualang_tools/digital_filters/README.md
```diff
@@ -0,0 +1,94 @@
...
+from the measured distortions:
+* [Single exponential correction](#singleexponentialcorrection): correct for a low-pass exponential decay `1 + A * exp(-t/tau)`.
+* [Highpass correction](#highpasscorrection): correct for a high pass exponential decay `exp(-t/tau)`.
+* [Bounce and delay correction](#bounceanddelaycorrection): correct for reflections and delays.
+* [Calc filter taps](#calcfiltertaps): correct for any combination of the aforementioned compensations.
+
+## Usage examples
+
+### single_exponential_correction
+Calculate the best FIR and IIR filter taps to correct for an exponential decay (undershoot and overshoot) of the shape
```
yomach, line 19:
```suggestion
Calculate the best FIR and IIR filter taps to correct for an exponential decay (undershoot or overshoot) of the shape
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - octave_tools - Added `get_calibration_parameters_from_db` get the most up-to-date correction parameters in the calibration database for the specified values of the Octave LO frequency, intermediate frequency and Octave gain.
 - octave_tools - Added `set_correction_parameters_to_opx` set the most up-to-date correction parameters from the calibration database for the specified values of the Octave LO frequency, intermediate frequency and Octave gain.
 - octave_tools - Added `get_correction_for_each_LO_and_IF` get the correction matrix elements for a set of intermediate frequencies picked equally spaced in a given list in order to update the correction matrix while performing the IF sweep in QUA.
+- digital_filters - Added a set of functions allowing the derivation the digital filter taps used to correct specific distortions.
```
yomach, line 15:
```suggestion
- digital_filters - Added library of functions allowing the derivation of the digital filter taps to correct distortions.
```

yomach (commented):
Looks good so far!

qualang_tools/digital_filters/README.md
TheoLaudatQM: done

CHANGELOG.md
TheoLaudatQM: done

TheoLaudatQM:
I fixed the function as we discussed and added some tests.
I have a question regarding the delay: you said: 
> if `delay` is not a multiple of the sampling rate, multiple FIR taps will be  created.
But even with delay=11 and Ts=2, I get 111 taps resulting in an error so I wanted to raise an error if delay is not a multiple of the sampling time, what do you think?

yomach:
> I fixed the function as we discussed and added some tests. I have a question regarding the delay: you said:
> 
> > if `delay` is not a multiple of the sampling rate, multiple FIR taps will be  created.
> 
> But even with delay=11 and Ts=2, I get 111 taps resulting in an error so I wanted to raise an error if delay is not a multiple of the sampling time, what do you think?

Ts = 2 is not a real scenario 😅 (Should be Ts = 0.5 for the OPX1000, in the future).
The relevant usecase, right now, is Ts=1 and delay=10.25.
We shouldn't raise an error, but we should allow trimming the number of taps (as part of applying the hardware limitation).
Something such as:
```
max_fir_taps = 44 - 7*len(feedback_taps)
if len(feedforward_taps)>max_fir_taps :
    feedforward_taps = feedforward_taps[:max_fir_taps ]
```

We can also add a message about it. I'm thinking something such as: summing the abs of the discarded FIR taps. If it's more than 0.01 (1%), give a warning, otherwise, just a debug message.
This is relevant because doing delay=1.5 would not really work (discarded taps will be >1%, probably ~10%), but delay=10.5 should be ok.

yomach:
I think that the errors/warning I suggested above would cover this "crap signal" scenario, WDYT?
We can discuss the hardware limitation.

qualang_tools/digital_filters/filters.py
```diff
@@ -2,19 +2,30 @@
...
+from warnings import warn
+from enum import Enum
+
+
+class QOP_VERSION(Enum):
+    QOP222 = {
+        "feedforward_max": 2 - 2**-16,
+        "feedback_max": 1 - 2**-20,
+        "feedforward_length": lambda feedback_len: 44 - 7 * feedback_len,
+    }
```
yomach, line 14: Looks good! I like the approach.
I would add another enum called "None" which has no limitations

qualang_tools/digital_filters/filters.py
```diff
@@ -52,12 +63,9 @@ def calc_filter_taps(
     if bounce is not None or delay is not None:
         feedforward_taps = bounce_and_delay_correction(bounce, delay, feedforward_taps, Ts)
 
-    max_value = max(np.abs(feedforward_taps))
-
-    if max_value >= 2:
-        feedforward_taps = 1.5 * feedforward_taps / max_value
+    feedforward_taps = _check_hardware_limitation_for_delay(qop_version, feedforward_taps, len(feedback_taps))
```
yomach, line 66: I would probably just combine it with the `_check_hardware_limitation` - Don't see a need for two functions and two calls

qualang_tools/digital_filters/filters.py
```diff
@@ -2,19 +2,30 @@
...
 
 def calc_filter_taps(
     fir: List[float] = None,
     exponential: List[Tuple[float, float]] = None,
     highpass: List[float] = None,
     bounce: List[Tuple[float, float]] = None,
-    delay: int = None,
+    delay: float = None,
     Ts: float = 1,
+    qop_version: Enum = QOP_VERSION.QOP222,
```
yomach, line 24: You're not passing these down to the other functions

qualang_tools/digital_filters/filters.py
```diff
@@ -81,9 +89,9 @@ def high_pass_exponential(x, t):
     return np.exp(-x / t)
 
 
-def single_exponential_correction(A: float, tau: float, Ts: float = 1):
+def single_exponential_correction(A: float, tau: float, Ts: float = 1, qop_version: Enum = QOP_VERSION.QOP222):
```
yomach, line 92: I would keep the default to all functions (besides the `calc` one above) at `qop_version: None`.
If we do want to keep this as is, then I would add another constant above:
`LATEST_QOP_VERSION =  QOP_VERSION.QOP222` and use it everywhere.

qualang_tools/digital_filters/filters.py
```diff
@@ -213,3 +228,30 @@ def _get_coefficients_for_delay(tau, full_taps_x, Ts=1.0):
...
+        warnings.formatwarning = _warning_on_one_line
+        warn(
+            f"The feedforward taps reached the maximum value of 2. \nThe coefficients are scaled down to stay within the valid range which reduces the outputted amplitude of the pulses played through the filtered port by a factor of {max_value/1.99:.3f}."
+        )
+    return list(feedforward_taps), list(feedback_taps)
+
+
+def _check_hardware_limitation_for_delay(qop_version: Enum, feedforward_taps: List, feedback_taps_len: int):
+    max_feedforward_len = qop_version.value["feedforward_length"](feedback_taps_len)
+    if len(feedforward_taps) > max_feedforward_len:
```
yomach, line 255: I would add a warning / error here according to what I wrote previously

qualang_tools/digital_filters/filters.py
```diff
@@ -213,3 +228,30 @@ def _get_coefficients_for_delay(tau, full_taps_x, Ts=1.0):
 def _round_taps_close_to_zero(taps, accuracy=1e-6):
     taps[np.abs(taps) < accuracy] = 0
     return taps
+
+
+def _check_hardware_limitation(qop_version: Enum, feedforward_taps: List, feedback_taps: List):
+    feedforward_taps = np.array(feedforward_taps)
+    feedback_taps = np.array(feedback_taps)
+    feedback_taps[feedback_taps > qop_version.value["feedback_max"]] = qop_version.value["feedback_max"]
```
yomach, line 236: I would also add a check here if this is used, and if it does, output a warning

qualang_tools/digital_filters/filters.py
TheoLaudatQM: What do you mean?
Here I just call it at the end so that the final taps are checked.
I added it to the other functions definition in case people directly get the taps from them without calling `calc_filter_taps()`.

qualang_tools/digital_filters/filters.py
TheoLaudatQM: I split it because I thought about calling the "delay limitation" only for the bounce and delay function, but it won't do anything to the other functions so I can combine them

qualang_tools/digital_filters/filters.py
TheoLaudatQM: Never I got what you meant and updated it :)

TheoLaudatQM: I set it to NONE for the other function so that the check is only performed at the return

qualang_tools/digital_filters/filters.py
TheoLaudatQM: My preference goes towards setting the latest QOP version as the default one, otherwise people will not specify it and it may lead to very bad results. For instance if the feedback tap is not capped and it is set to 1, then the OPX will output garbage data from -0.5V to +0.5V which can kill some spin qubit devices (it almost happened to me during an installation). I like your LATEST_QOP_VERSION  idea!

TheoLaudatQM:
If we raise an error, then it will be solved, but if we only raise a warning, then the user will be notified about it but the OPX will still output something that can be harmful for the device...
Maybe I can add a `safety` option set to True by default that will raise an error and the user can set it to False if he still wants to output the signal. WDYT?

TheoLaudatQM (approved):

PR #178, opened 2023-12-08
Modifications to callable_from_qua

nulinspiratie:
This is a PR attempting to implement the changes required for callable_from_qua.
@TheoLaudatQM I'm not able to run this locally, so there are likely to be several bugs. can you try this out nonetheless?

Following our earlier conversation with @TheoLaudatQM @yonatanrqm @yomach I'll try to implement the changes related to inheriting from `Program` and `QuantumMachine`.

The goal is that we no longer need to run the following:
```python
with prog.local_run(qm):
    job = qm.execute(prog)
```
but can instead directly run
```python
job = qm.execute(prog)
```

Usage:
```python
from qualang_tools.callable_from_qua import *
patch_callable_from_qua()  # Patch qua classes
enable_callable_from_qua()  # Attach callable_from_qua addon

@callable_from_qua
def execute_function():
    print("I'm being run locally")

with program() as prog:
   execute_function()

qmm = QuantumMachinesManager(...)
config = ...
qm = qmm.open_qm(config)
qm.execute(program).
```

I think the following items should be done
- [x] Inherit from `QuantumMachine`, add `QuantumMachine.callables` which get executed when calling `qm.execute()`
- [x] Patch `Program`, add `program.addons`
- [x] Rename `@local_run` back to `@callable_from_qua`
- [x] Rename `LocalRunManager` to `QuaCallableEventManager`
- [x] Move `callable_from_qua` away from `qualang_tools/addons` to `qualang_tools/callable_from_qua`. @yomach do you agree

Open questions
- Should the `CallableEventManager` be opened in the main thread or in a separate thread while the program is executing on the machine?   
  I think the best is if it by default opens in a separate thread, but the user can specify to run it in the main thread, in which case it's blocking.
- How to pass additional callables?  
  E.g. we can currently run `with prog.local_run(qm, funcs=[live_plot]):` to also add live plotting


yomach:
I agree with what you wrote (haven't looked at the code yet).
@TheoLaudatQM - Note that this is a PR to merge to **your** branch, not to main.
So we please test it locally and if it's good, we'll merge it there and continue development from there.

qualang_tools/callable_from_qua/_qua_patches.py
```diff
@@ -0,0 +1,102 @@
...
+        program: Program,
+        duration_limit: int = 1000,
+        data_limit: int = 20000,
+        force_execution: int = False,
+        dry_run: int = False,
+        simulate: Optional[SimulationConfig] = None,
+        compiler_options: Optional[CompilerOptionArguments] = None,
+    ) -> RunningQmJob:
+        for program_addon in program.addons.values():
+            program_addon.execute_program(program=program, quantum_machine=self)
```
TheoLaudatQM, line 68: In execute_program(), you're trying to get the running job `job = quantum_machine.get_running_job()` which doesn't already exist since it will be created in the next line. This prevents the program from running and I am not sure how to fix it myself...

TheoLaudatQM (changes requested):
The test programs don't run because a running job is needed before it is actually created

TheoLaudatQM (approved):

PR #180, opened 2023-12-18
Video Mode

TheoLaudatQM:
This module allows the user to update some pre-defined parameters of a QUA program while fetching data from the OPX for dynamic tuning (quantum dots, PID loop,...).


examples/video_mode/test_PID.py
```diff
@@ -0,0 +1,186 @@
+from matplotlib import pyplot as plt
```
yomach, line 1: names starting with `test_` are commonly used for testing with PyTests. Let's rename to PID_example? or just PID?

same for user input

qualang_tools/video_mode/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+Alternatively, you can also access the QUA variables later in the code as follows:
+```
+with program() as video_mode_prog:
+    amp_param1, amp_param2, ... = video_mode.declare_variables()
+    
+    with infinite_loop():
+        video_mode.load_parameters()
+        
+        play("my_pulse"*amp(amp_param1, "qe1")
```
yomach, line 130:
```suggestion
        play("my_pulse"*amp(amp_param1), "qe1")
```

qualang_tools/video_mode/README.md
```diff
@@ -0,0 +1,184 @@
...
+        play("my_pulse"*amp(video_mode["amp_param"]), "qe1")
+        play("my_other_pulse"*amp(video_mode["amp_param"]), "qe2")
+```
+
+Note that you can also access to the full list of QUA variables declared in the program through the ```video_mode.variables``` attribute.
+
+Alternatively, you can also access the QUA variables later in the code as follows:
+```
+with program() as video_mode_prog:
+    amp_param1, amp_param2, ... = video_mode.declare_variables()
```
yomach, line 125: Is the ordered guaranteed according to the dict order?
I would guess so, but it needs to be stated

examples/video_mode/test_user_input.py
```diff
@@ -0,0 +1,80 @@
...
+    param_dict = {
+        "dc_offset_1": 0.0,
+        "dc_offset_2": 0.0,
+    }
+    # Initialize the video mode
+    video_mode = VideoMode(qm, param_dict)
+    # Get the QUA program
+    qua_prog = qua_prog(video_mode)
+    # Execute the QUA program in video mode
+    job = video_mode.execute(qua_prog)
```
yomach, line 61: I don't see anywhere in the code that the parameters are being updated, don't we want to show that?

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,444 @@
...
+            "'param_name': returns the value of the parameter.\n"
+        )
+
+        signal.signal(signal.SIGINT, self.signal_handler)
+        signal.signal(signal.SIGTERM, self.signal_handler)
+
+    def signal_handler(self, signum, frame):
+        """Signal handler for SIGTERM and SIGINT signals."""
+        print(f"Received signal {signum}, stopping VideoMode...")
+        self.qm.set_io1_value(666)
```
yomach, line 237: why 666?

If it's a magic number, then change it from 666, and put it in a constant variable.

qualang_tools/video_mode/README.md
```diff
@@ -0,0 +1,184 @@
...
+        play("my_pulse"*amp(amp_), "qe1")
+        frame_rotation(phase, "qe1")
+```
+Finally, we start the video mode execution:
+```
+video_mode.execute(video_mode_prog)
+```
+The user can then interact with the program by typing the following in the Python console:
+```
+amp=0.8
```
yomach, line 177: This is not 100% clear to me
Does it work when there isn't a console?

Any, it's missing from the examples...

examples/video_mode/test_user_input.py
arthurostrauss: The functionality of the VideoMode is that the parameters are loaded by the user through his terminal or the Python console (meaning he has to type them down once the program is already started). The main objective of this mode is to enable a change of parameters without knowing their values in advance while the program runs in the background. Everything happens therefore with user input() Python function which is handled behind the scenes in the class. There's no parameter update to show in the script since the user is summoned to update them once the program is launched (through print statements and waiting for input in the console, while the QUA program runs in the background)

qualang_tools/video_mode/videomode.py
arthurostrauss: The value is chosen such that the switch_case that is evaluated within the program could never encounter a case with the input IO value. It could be any large number which would not fit in a reasonable switch case for this particular use case

qualang_tools/video_mode/README.md
arthurostrauss: It works in a notebook (example is available there), in a terminal, and in the console. The whole idea is that the user should have the ability to type some input parameters with the keyboard while the program is running (like a prompt). Could you specify what you mean by missing from the examples?

qualang_tools/video_mode/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+In some cases, one might want to update parameters of the program while it is running and without knowing in advance when and how to update them.
+We present here the ```VideoMode``` which enables such dynamic parameter tuning.
+
+## How to use it
+
+The user can declare a ``VideoMode`` instance by passing in a ```QuantumMachine``` instance and a dictionary of
+parameters intended to be updated dynamically. Additionally, if a job is already started on the ```QuantumMachine```,
+it can also use this job to change dynamically some of its parameters (this assumes that the QUA program associated to this job
+was already designed to be compatible with the ```VideoMode```, which we will explain how to do below).
```
yomach, line 12: Is this explained?

qualang_tools/video_mode/README.md
```diff
@@ -0,0 +1,185 @@
...
+    with infinite_loop():
+        video_mode.load_parameters()
+        
+        play("my_pulse"*amp(amp_param1), "qe1")
+        play("my_other_pulse"*amp(amp_param2), "qe2")
+```
+## Outside the QUA program
+The ```VideoMode``` class has only one ```execute```method to be called outside the QUA program scope, which is a simple wrapper of ```qm.execute()``` method.
+It will start the video mode execution by creating a new thread doing a continuous query of input parameters from the user, while the program is executed and possible live data plotting is on.
+Note that the QUA program runs in the background and the user can still interact with the Python environment while the program is running.
```
yomach, line 137: This line is confusing, it sounds like the normal execute which let's you **freely** interact with python. But if I understood it correctly, you've taken over the console?

This needs rephrasing. I think this is what confused me about the examples

qualang_tools/video_mode/README.md
yomach: The notebook wasn't around when I wrote this.

yomach (changes requested):

examples/video_mode/test_user_input.py
yomach: Ok, I know understand you take over the console.
This isn't explained anywhere, but should appear in the readme and not here

qualang_tools/video_mode/videomode.py
yomach: I would go for `-1`, but you can go for `2^31-1`.
Then as I wrote, change it to something nobody will find offending.
In any case, it's a magic number, so put it in a constant variable:
`EMPTY_OPTION = 2**31-1`

qualang_tools/video_mode/README.md
yomach: I would add: 
At this point, the python console accepts specific inputs, type `help` to see the full list. For example, to change the variables, type:

qualang_tools/video_mode/README.md
arthurostrauss: Let me know if the new sentence clarifies it.

qualang_tools/video_mode/README.md
arthurostrauss: This features is now removed as it eventually did not make sense to support it for the use case at hand.

qualang_tools/video_mode/README.md
yomach: Yes it does, thanks!

yomach:
So, I haven't gone through the actual code too much. I suggest we ask Rosmarin for a review, and also present it to a few other CS people to comment on small things such as readme and prompts.
But it looks very nice!

TheoLaudatQM:
Thanks, we already showed it to the CS team, but I kind of already tested it with Yiwen Chu students (PID example) and they liked it. The goal is to keep it confidential at the beginning so that we can test it with a few customers, the goal being to have it working for tuning up Quantum dots devices.
I am fine with Rosmarin reviewing it, but since Niv put it in the OKRs, it needs to be merged by tomorrow...

yomach:
> Thanks, we already showed it to the CS team, but I kind of already tested it with Yiwen Chu students (PID example) and they liked it. The goal is to keep it confidential at the beginning so that we can test it with a few customers, the goal being to have it working for tuning up Quantum dots devices. I am fine with Rosmarin reviewing it, but since Niv put it in the OKRs, it needs to be merged by tomorrow...

Then talk with Rosmarin and make sure he reviews it in time :)

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+                if float(parameter_value).is_integer() and parameter_value > 8:
+                    self.table[parameter_name][
+                        "declare_expression"
+                    ] = f"declare(int, value=int({parameter_value}))"
+                    self.table[parameter_name]["type"] = int
+                else:
+                    self.table[parameter_name][
+                        "declare_expression"
+                    ] = f"declare(fixed, value={parameter_value})"
+                    self.table[parameter_name]["type"] = float
```
yonatanrqm, line 46: Use a class for the declaration data, don't use dictionaries

yonatanrqm (changes requested):

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+            pause_program: Boolean indicating whether the program should be paused while waiting for user input.
+        """
+
+        if pause_program:
+            pause()
+
+        param_index_var = declare(int)
+        assign(param_index_var, IO1)
+
+        with switch_(param_index_var):
```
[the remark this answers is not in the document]
TheoLaudatQM: From what I know, input stream is blocking the execution and this is something we absolutely don't want. Here we need to asynchronously update the parameters of the running program...

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+                    "Invalid parameter type. Please use float, int or bool or list."
+                )
+
+    def declare_variables(self):
+        """
+        QUA Macro to create the QUA variables associated with the parameter table.
+        """
+        for parameter_name, parameter in self.table.items():
+            if parameter["declare_expression"] is not None:
+                exec(f"{parameter_name} = {parameter['declare_expression']}")
```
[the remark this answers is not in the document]
TheoLaudatQM: I will let @arthurostrauss answer this one, because I think `exec` was the key function that makes the tool work

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+                self.table[parameter_name]["var"] = eval(parameter_name)
+        pause()
+        if len(self.variables) == 1:
```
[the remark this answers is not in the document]
TheoLaudatQM: This would be annoying from the user standpoint because he is getting the variables as `var1, var2, var3 = declare_variables()`. If the function returns a list of one element, then the user will need to write `var1 = declare_variables()[0]` no?

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+from qm.qua import *
+from qm import QuantumMachine, QmJob, Program
+import time
+import signal
+from typing import Optional, List, Dict, Union
+from dataclasses import dataclass
+import numpy as np
+
+
+@dataclass
```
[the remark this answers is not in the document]
TheoLaudatQM: I will let @arthurostrauss comment

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+    parameters_dict: Dict[str, Union[float, int, bool, List, np.ndarray]]
+
+    def __post_init__(self):
+        self.table = {}
+        for index, (parameter_name, parameter_value) in enumerate(
+            self.parameters_dict.items()
+        ):
+            self.table[parameter_name] = {"index": index, "value": parameter_value}
+            if isinstance(parameter_value, float):
+                if float(parameter_value).is_integer() and parameter_value > 8:
```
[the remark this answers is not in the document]
TheoLaudatQM: good point, I fixed it

qualang_tools/video_mode/videomode.py
```diff
@@ -0,0 +1,452 @@
...
+                    self.table[parameter_name][
+                        "declare_expression"
+                    ] = f"declare(int, value=int({parameter_value}))"
```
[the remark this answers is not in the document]
TheoLaudatQM: Yes I understand your point, however the idea is to ask the minimum from the user so that it will be easy for him/her to implement and I had in mind that `var=0.0 --> fixed` and `var=0 --> int`.
We can also add the type of the QUA variable as a key of `param_dict`, what do you think? @arthurostrauss ?

qualang_tools/video_mode/videomode.py
TheoLaudatQM: Do you mean for `param_dict`?

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
yomach: What Theo wrote.
In the future, the input stream will support this capability...

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
yomach: I agree with Yonatan on this one. I think that declaring the type is not a large overhead

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
arthurostrauss: The whole point of the class is to conceal the complexity of usage with natural calls. Although I would recommend the user to retrieve the QUA variables in the program through `video_mode["param_name"]` and not by declaring them as shown in Théo's version, but I figure it is a matter of taste. I would leave it as it is to enable both versions.

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
arthurostrauss: Got it, I removed the data class and created a usual class instead

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
arthurostrauss: So you would like the input param_dict to be of the form:
`{param_name:
{"value": initial_value,
"type": type(initial_value)}
}`

Is that correct?  
The overhead to declare the type of each variable attached to each parameter in the input dict is in my opinion an unnecessary redundancy, as if the user has lots of parameters, all initialised to Python values, he will in most cases want to do type(parameter_value) anyway. The error appears later in the code if he tries to update an int with a float variable for example. 

I think the extra attention to let the user know that he should define carefully the initial_value is enough to avoid the explicit type definition of the variable type (especially because we have a limited set of such types).

qualang_tools/video_mode/videomode.py
arthurostrauss: Confusing for me too, you would like to define a class for each single parameter with the attributes corresponding to var, type, index ?

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
TheoLaudatQM: I think that if we make the user specify the type, then it must be a "QUA" type where the only possibilities are `int`, `bool` or `fixed`, no matter the initial values. It is true that now, if you initialize a variable with 0, then it will be declared as an `int`, which may not be what the user wants...
I the majority thinks it is better (more robust and less confusing for the user) to specify the type, then I am also for it

yomach (approved):
I think it's good.
Just need to finalize whether the user declares the type or not

qualang_tools/video_mode/videomode.py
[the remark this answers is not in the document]
arthurostrauss: I suggest forcing the type as being something optional. That means that we maintain compatibility with the current inferred type as done right now, and on top of it, if the user specifies the type we do an additional check that the value matches the provided input type to declare the QUA variable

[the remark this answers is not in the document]
yomach: I'm against automatic inferring of the variable type, as I believe it will cause errors, and the error message will not be indicative enough: A user will get an error that duration cannot accept a fixed point parameter, but he will have no way of knowing how to fix it. Example of error message:
`2023-12-24 13:28:20,824 - qm - ERROR    - Expected type int but got fixed (@ File "D:\GitHub\qm-cs-yoav\IQM\IQ_blobs.py", line 361: play("pi_half", "q1_xy", duration=I) )`

Moreover, users are used to declaring the variable type.

But with that being said, I'm leaving this up to CS to decide, so if both @TheoLaudatQM and @arthurostrauss want to automatically infer it, go ahead.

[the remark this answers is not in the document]
arthurostrauss: I modified the code to take into account the forced type specification. I left available the possibility to infer it automatically but I moved it away from the documentation for now. All examples and tests are updated accordingly

PR #182, opened 2023-12-21
octave_tools

TheoLaudatQM:
Functions to ease the update of the correction parameters when performing wide frequency sweeps (LO sweep in Python and IF in QUA)


qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
+# Octave Tools
+This library includes tools to improve the users Octave experience.
+
+**Note that these functions only work with `qm-qua >= 1.1.5`.**
```
yomach, line 4: Since the dependency is `qm-qua == 1.1.7`, I don't think we need to mention it

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+# Calibrate the desired set of LO frequency and IF
+qm.octave.set_rf_output_gain("qubit", -7)
+qm.calibrate_element("qubit", {5e9: (50e6,)})
+
+# Set the new frequency with calibrated gain
+qm.octave.set_lo_frequency("qubit", 5e9)
+qm.octave.set_rf_output_gain("qubit", -7)
+# Get the correction parameters corresponding to the desired parameters:
+param_qubit = set_correction_parameters(
```
yomach, line 39: The example here is for set

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+# Open the QMM
+qmm = QuantumMachinesManager()
+# Open the quantum machine
+qm = qmm.open_qm(config)
+# Loop over the LO frequencies
+for i in range(len(lo_frequencies)):  
+    # Set the frequency and gain of the LO source
+    qm.octave.set_lo_frequency("qubit", lo_frequencies[i])
+    qm.octave.set_rf_output_gain("qubit", 0)
```
yomach, line 88: Maybe better not to set the gain every time?

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+    # Update the correction parameters
+    set_correction_parameters(
```
yomach, line 90: Would `set_calibration_db(get_calibration_db(...))` also work? It seems cleaner to me, but I'm not sure...

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+    calibrate=True,
+    qm=qm,
+)
+
+with program() as LO_sweep_prog:
+    # QUA declarations
+    c00_qua = declare(fixed, value=c00)  # QUA variable for the measured 'I' quadrature
+    c01_qua = declare(fixed, value=c01)  # QUA variable for the measured 'I' quadrature
+    c10_qua = declare(fixed, value=c10)  # QUA variable for the measured 'I' quadrature
+    c11_qua = declare(fixed, value=c11)  # QUA variable for the measured 'I' quadrature
```
yomach, line 163: comments need to be removed

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+intermediate_frequencies = np.arange(f_min, f_max + 0.1, df)
+# The LO frequency sweep parameters
+f_min_external = 5.001e9 - f_min
+f_max_external = 5.5e9 - f_max
+df_external = f_max - f_min
+lo_frequencies = np.arange(f_min_external, f_max_external + df_external / 2, df_external)
+
+# Get the list of intermediate frequencies at which the correction matrix will 
+# be updated in QUA and the corresponding correction matrix elements
+IFs, c00, c01, c10, c11 = update_correction_for_each_IF(
```
yomach, line 146: what about the LOs? how would that work when there are multiple LOs?

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+correction matrix) corresponding to a given set of intermediate frequency, Octave LO frequency and gain. 
+
+The goal to give the user the ability to easily update the correction parameters while sweeping the LO frequency
+in Python for instance.
+
+The correction parameters are returned in the format of a dictionary:
+```
+{'offsets': {'I': I_offset, 'Q': Q_offset},
+ 'correction_matrix': (c00, c01, c10, c11)}
+```
```
yomach, line 56: remove?

qualang_tools/octave_tools/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+If no calibration parameters are found in the database, the function will either:
+* return the dictionary with empty values (verbose_level 0)
+* return the dictionary with empty values and print a warning in the Python console (verbose_level 1)
+* raise an error and block the execution of the script (verbose_level 2)
```
yomach, line 61: Need to update wording: It will not "return the dictionary", but it will "not update the parameters"?

qualang_tools/octave_tools/octave_tools.py
```diff
@@ -0,0 +1,199 @@
...
+    element: str,
+    LO: float,
+    IF: float,
+    gain: float,
+    qm: QuantumMachine,
+    job: Union[QmJob, RunningQmJob] = None,
+    verbose_level: int = 2,
+) -> dict:
+    """Look for the correction parameters in the database, located at the specified path_to_database, for the specified values of
+    the Octave LO frequency, intermediate frequency and Octave gain and update the running job.
```
yomach, line 85: Need to mention that if no job is specified, it automatically tries to take the currently running job

qualang_tools/octave_tools/octave_tools.py
```diff
@@ -0,0 +1,199 @@
...
+    The function will return the list on intermediate frequencies at which the correction matrix will be updated in the
+    program, the 'I' and 'Q' offsets and the four coefficients of the correction matrix with one element for each pair (LO, IF).
+
+    :param path_to_database: direct path to the calibration database (without '/calibration_db.json').
+    :param config: the OPX config dictionary.
+    :param element: the element to get the correction parameters for.
+    :param gain: the Octave gain.
+    :param LO_list: list of the LO frequencies involved in the scan, in Hz.
+    :param IF_list: list of the intermediate frequencies involved in the scan, in Hz.
+    :param nb_of_updates: number of intermediate frequencies to calibrate and for which the program will update the correction pmatrix.
```
yomach, line 144: I think I understand it, but maybe it needs more explaining:
If you give a list of 100 IFs, but nb_of_updates is 10, then it will only calibrate on every 10th frequency? Maybe an example?

yomach (commented):
I did zero testing, just went over readme + changelog + docstrings.
Attached are comments.
I suggest you also have people (CS?) play test it first

qualang_tools/octave_tools/README.md
TheoLaudatQM: Good point, it is not needed anymore

qualang_tools/octave_tools/README.md
TheoLaudatQM: Not as it is, do you think it is better to split `set` and `get` rather than having `get` and `get+set`?

qualang_tools/octave_tools/README.md
TheoLaudatQM: Yes it works for a list of LOs and IFs, I changed the name to make it clearer

qualang_tools/octave_tools/README.md
TheoLaudatQM: Why? Here I wanted to explicitly show the structure

qualang_tools/octave_tools/octave_tools.py
TheoLaudatQM: Good point, yes your understanding is correct, I added the example

qualang_tools/octave_tools/README.md
yomach: Never mind. I wrote it before I understood how it works, The `set` automatically calls the `get`.
You can ignore and resolve

qualang_tools/octave_tools/README.md
yomach: I think what is confusing is the resulting lengths:
IF : Length is `nb_update_freqs`
cxx: Length is `nb_update_freqs*len(LO)`

qualang_tools/octave_tools/README.md
yomach: This is in the set function (I agree it's needed in the get)

TheoLaudatQM: Actually the set function also returns the calibration parameters in case the user wants to check it, this is why I kept it here as well

yomach: Oh. Then it should be stated... Also, it doesn't appear in the docstring.

I understand what bothers me. Usually, you have a set and a get function to the same "object".
Here, the set and the get are for different objects: The get is from the calibration, the set is to the OPX.
That's ok, just needs to be explained. And this is actually clear in the function names and docstrings, so I wonder if we're already good?
The functions names are actually different, but "calibration" and "correction" are hard to distinguish from a quick glance...

TheoLaudatQM: Yes I just stated it in the returns, I added it explicitly in the docstring now.
I wanted to make the users use only one function depending whether they want to get or set the parameters, but maybe it is just confusing...
If you think it is better I can keep the `get` function as it is and modify the `set` function to only set the parameters from what the `get` function returns as you suggested initially, what do you think?

qualang_tools/octave_tools/README.md
TheoLaudatQM: Indeed, I specified it in the readme and docstring

qualang_tools/octave_tools/README.md
yomach: I'm not sure... Maybe run it with a few CS people and see what makes sense to them.
To me, having the `set` get the parameters as input is better, but it's also confusing because it updates something differently from the `get` 😆 

Maybe `get_calibration_parameters_from_db` & `set_correction_parameters_to_opx`?

TheoLaudatQM: The problem is that the bandwidth is super limited because of AWS...
I like your suggestion for the names, it is way more explicit!
However, there is no straightforward way to set to correction matrix if a job is not running, right?

yomach: Yes, I understand...
Maybe the name change is enough. We can change it to `set_correction_parameters_to_job`.
Do we only intend it to be used on a running job?

TheoLaudatQM: I would prefer not to, but the only available function I could find for loading the correction matrix is from the job API...

yomach: `qm.set_mixer_correction()`

TheoLaudatQM: Yeah but you need to find the mixer name which is hidden with the Octave and maybe be subject to breaking changes...
I started to do this and then it looked to hacky for me and potentially not very robust because if you have several elements in the mixer list, then you need to make sure to modify the right one...
Do you think I should still do it?

yomach: We can give it an element, get the config from the QM, then get the mixer name from the element.

And no, I think that doing it only for a job is ok, let's just change the function name.

TheoLaudatQM: After some thinking I realized that several customers are changing the LO frequency using the API within the program and not the config and they want to update the correction parameters before running a job.
I implemented the possibility to do it, but since `qm.set_mixer_correction()` only adds a set of parameters in the `mixers` section, you also need to update the LO and intermediate frequencies of the element so that the OPX picks the right set of correction parameters.
I didn't manage to do so because the qm API doesn't allow updating the frequencies in the loaded config (without reopening a QM) so I just updated the correction matrix but not the corresponding frequencies (this is what I called hacky above). For instance:
`{'intermediate_frequency': 50e6, 'lo_frequency': 4e9, 'correction': old_matrix)}`
`set_correction_parameters_to_opx(if=-10e6, lo=6e9)`
`{'intermediate_frequency': 50000000.0, 'lo_frequency': 4000000000.0, 'correction': new_matrix)}`
What do you think?

yomach: I think that this is the correct solution for now.
In the (near) future, the correction parameters will not be part of the QM, but part of the job.

yomach: What about changing the names of the functions?
`get_calibration_parameters_from_db` & `set_correction_parameters_to_opx`

TheoLaudatQM: I already updated the names, do you mean changing these to something else?

yomach: Oh, I missed that.
Please update it in the changelog as well!

TheoLaudatQM: Good catch, done!

yomach (approved):

PR #183, opened 2024-01-02
simulator tools -> better create connections

yomach:
This creates the controller connections in a much nicer way, which also matches the documenation.
I think we can completely remove the `old_method`.
It also adds an option to use a subset of the controllers, but currently only remove those from the end (so if you have a 6 OPX setup, you can use 1-4, but not 3-6), but that can be edited.
WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+### Added
+- simulator_tools - ``create_simulator_controller_connections`` can now be used to create the connections between a subset of a large cluster.
+
 ### Changed
 - config/waveform_tools - Added sampling rate argument with default value set to 1GS/s to the waveforms.
+- simulator_tools - ``create_simulator_controller_connections`` now creates the connections with a different algorithm that uses all available optical connections.
+- simulator_tools - ``create_simulator_controller_connections`` order of input parameters has changed.
+
+### Deprecated
+- simulator_tools - ``qualang_tools.simulator_tools`` has been deprecated and was moved to ``qualang_tools.simulator``.
 
 ## [0.16.0] - 2024-01-25
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
-- simulator_tools - ``create_simulator_controller_connections`` can now be used to create the connections between a subset of a large cluster.
+- simulator - ``create_simulator_controller_connections`` can now be used to create the connections between a subset of a large cluster.
 
 ### Changed
 - config/waveform_tools - Added sampling rate argument with default value set to 1GS/s to the waveforms.
-- simulator_tools - ``create_simulator_controller_connections`` now creates the connections with a different algorithm that uses all available optical connections.
-- simulator_tools - ``create_simulator_controller_connections`` order of input parameters has changed.
+- simulator - ``create_simulator_controller_connections`` now creates the connections with a different algorithm that uses all available optical connections.
+- simulator - ``create_simulator_controller_connections`` order of input parameters has changed.
 
 ### Deprecated
-- simulator_tools - ``qualang_tools.simulator_tools`` has been deprecated and was moved to ``qualang_tools.simulator``.
+- simulator - ``qualang_tools.simulator_tools`` has been deprecated and was moved to ``qualang_tools.simulator``.
 
 ## [0.16.0] - 2024-01-25
 ### Fixed
--- a/qualang_tools/results/README.md
+++ b/qualang_tools/results/README.md
@@ -20,7 +20,7 @@ n_avg = 1000
 with program() as prog:
     # QUA program with n_avg averaging iterations
 
-qmm = QuantumMachinesManager(qop_ip, cluster_name)
+qmm = QuantumMachinesManager(host=qop_ip, port=qop_port, cluster_name=cluster_name)
 qm = qmm.open_qm(config)
 job = qm.execute(prog)
 
--- a/qualang_tools/results/README.md
+++ b/qualang_tools/results/README.md
@@ -17,10 +17,10 @@ Then the results can be fetched with the `.fetch_all()` method while the program
 from qualang_tools.results import fetching_tool
 
 n_avg = 1000
-with program as prog:
+with program() as prog:
     # QUA program with n_avg averaging iterations
 
-qmm = QuantumMachinesManager(host="127.0.0.1", port="80")
+qmm = QuantumMachinesManager(qop_ip, cluster_name)
 qm = qmm.open_qm(config)
 job = qm.execute(prog)
 
--- a/qualang_tools/simulator/README.md
+++ b/qualang_tools/simulator/README.md
@@ -3,8 +3,8 @@ This library includes tools to help simulate programs
 
 ## create_simulator_controller_connections
 Creates a list of :class:`~qm.simulate.interface.ControllerConnection` objects with the Inter-controllers connections between controllers in a way which maximizes the optical links connectivity.
-For use inside the :class:`~qm.simulate.interface.SimulationConfig` for simulating the controllers` behavior.
-Can be used to also simulate part of a small cluster: For example, the following command would create all the connections
+For use inside the :class:`~qm.simulate.interface.SimulationConfig` for simulating the controllers' behavior.
+It can be used to also simulate part of a small cluster: For example, the following command would create all the connections
 needed to run a simulation on 'con1', 'con4', 'con5' which are part of a 9 OPX cluster: `create_simulator_controller_connections(9, [1,4,5])`
 
 ### Usage example
@@ -15,7 +15,7 @@ from qualang_tools.simulator import create_simulator_controller_connections
 with program() as prog:
     # QUA program 
 
-qmm = QuantumMachinesManager(qop_ip, cluster_name)
+qmm = QuantumMachinesManager(host=qop_ip, port=qop_port, cluster_name=cluster_name)
 job = qmm.simulate(config,
                    prog,
                    SimulationConfig(simulation_duration,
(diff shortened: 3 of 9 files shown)


qualang_tools/simulator_tools.py
```diff
@@ -15,35 +16,66 @@ def create_simulator_controller_connections(n_controllers, n_connections=12):
     :returns: The ControllerConnection object
```
TheoLaudatQM, line 16: The new parameters are missing from the docstring.
Are there some case for keeping using the old method? If not, then shouldn't we had a depreciation warning to transition towards the "new" method?

qualang_tools/simulator_tools.py
```diff
@@ -1,7 +1,8 @@
+import numpy as np
 from qm import ControllerConnection, InterOpxChannel
 
 
-def create_simulator_controller_connections(n_controllers, n_connections=12):
+def create_simulator_controller_connections(n_controllers, n_connections=12, old_method=False, n_actual_controllers=None):
```
TheoLaudatQM, line 5: If I understand correctly, n_controllers=6 and n_actual_controller=4 would mean that you want to "build" only a subset of the 6 controllers and it will be from 1 to 4, is it correct?
What is the use-case for this?

TheoLaudatQM (changes requested):
I actually never use this tools so I am not very familiar with the use-cases.
In any case the readme and docstring need to be updated and maybe also have some unit tests. Writing this I actually realized that there is no readme for this tool...

qualang_tools/simulator_tools.py
yomach: Good points!

qualang_tools/simulator_tools.py
yomach: Yes, exactly.
It came from TII, which has a 9 OPX cluster connected but then opened a 3 OPX cluster. So the topology was for 9, but the cluster was for 3.
Maybe we can change the naming a bit?

Also, as Tomer pointed out to me recently, this doesn't cover all of the use cases, as sometimes you need devices 4-6 and not 1-3 (which is what is given here).

Any suggestions on the syntax for that?

yomach:
> I actually never use this tools so I am not very familiar with the use-cases. In any case the readme and docstring need to be updated and maybe also have some unit tests. Writing this I actually realized that there is no readme for this tool...

Thanks!
This is a very old branch, and I never got around to writing the unit tests :\ Which is probably the only reason it wasn't merged.
But Tomer recently needed it, so I came back to it.
I'll try to finalize this (and also the readme!)

qualang_tools/simulator_tools.py
TheoLaudatQM: I see, thanks!
Then maybe something like subsets=(4, 5, 6) could work no? Like this you can also define a subset as non-neighbors like (1, 5, 6) for instance, although I don't know if it makes sense...

qualang_tools/simulator_tools.py
yomach: I decided just to remove the old method

qualang_tools/simulator_tools.py
yomach: Done

qualang_tools/results/README.md
```diff
@@ -17,10 +17,10 @@ Then the results can be fetched with the `.fetch_all()` method while the program
 from qualang_tools.results import fetching_tool
 
 n_avg = 1000
-with program as prog:
+with program() as prog:
     # QUA program with n_avg averaging iterations
 
-qmm = QuantumMachinesManager(host="127.0.0.1", port="80")
+qmm = QuantumMachinesManager(qop_ip, cluster_name)
```
TheoLaudatQM, line 23: The second argument of `QuantumMachinesManager` is still the port so I don't think this would work...
Maybe we can keep `qmm = QuantumMachinesManager(host=qop_ip, port=qop_port, cluster_name=cluster_name)`

qualang_tools/simulator_tools.py
```diff
@@ -1,49 +1,14 @@
...
+__all__ = [
+    "create_simulator_controller_connections",
+]
+
+warnings.warn(
+    "The 'create_simulator_controller_connections' function has been moved to 'qualang_tools.simulator'.", FutureWarning
+)
+warnings.warn(
+    "The 'create_simulator_controller_connections' function has been moved to 'qualang_tools.simulator'.",
+    DeprecationWarning,
```
TheoLaudatQM, line 13: Why do you have `FutureWarnign` and `DepreciationWarning` here and not ust one of them?

qualang_tools/simulator/README.md
```diff
@@ -0,0 +1,43 @@
+# Simulator tools
+This library includes tools to help simulate programs
+
+## create_simulator_controller_connections
+Creates a list of :class:`~qm.simulate.interface.ControllerConnection` objects with the Inter-controllers connections between controllers in a way which maximizes the optical links connectivity.
+For use inside the :class:`~qm.simulate.interface.SimulationConfig` for simulating the controllers` behavior.
```
TheoLaudatQM, line 6: I think ` should be ' after controllers

qualang_tools/simulator/README.md
```diff
@@ -0,0 +1,43 @@
...
+Can be used to also simulate part of a small cluster: For example, the following command would create all the connections
```
TheoLaudatQM, line 7: **It** can be used

qualang_tools/simulator/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Usage example
+A simple example for a 3 OPX cluster:
+```python
+from qualang_tools.simulator import create_simulator_controller_connections
+
+with program() as prog:
+    # QUA program 
+
+qmm = QuantumMachinesManager(qop_ip, cluster_name)
```
TheoLaudatQM, line 18: If you want to show an example with cluster name, then the keyword must be specified:
`qmm = QuantumMachinesManager(qop_ip, cluster_name=cluster_name)` (at least with qm-qua==1.1.6)

qualang_tools/simulator/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+ 
+A simulation running on 'con1', 'con4', 'con5' which are part of a 9 OPX cluster, which also has loopback connected between the controllers:
+```python
+from qualang_tools.simulator import create_simulator_controller_connections
+
+with program() as prog:
+    # QUA program 
+
+qmm = QuantumMachinesManager(qop_ip, cluster_name)
```
TheoLaudatQM, line 32: same hee

CHANGELOG.md
```diff
@@ -4,8 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+### Added
+- simulator_tools - ``create_simulator_controller_connections`` can now be used to create the connections between a subset of a large cluster.
+
 ### Changed
 - config/waveform_tools - Added sampling rate argument with default value set to 1GS/s to the waveforms.
+- simulator_tools - ``create_simulator_controller_connections`` now creates the connections with a different algorithm that uses all available optical connections.
+- simulator_tools - ``create_simulator_controller_connections`` order of input parameters has changed.
```
TheoLaudatQM, line 13: The name is now `simulator` instead of `simulator_tool`s, no?

CHANGELOG.md
```diff
@@ -4,8 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+### Added
+- simulator_tools - ``create_simulator_controller_connections`` can now be used to create the connections between a subset of a large cluster.
```
TheoLaudatQM, line 8: The name is now simulator instead of simulator_tools, no?

qualang_tools/simulator/simulator.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+from qm import ControllerConnection, InterOpxChannel
+
+
+def create_simulator_controller_connections(
+    n_controllers, actual_controllers=None, n_connections=12, print_debug=False
+):
+    """
+    Creates a list of :class:`~qm.simulate.interface.ControllerConnection` objects with the Inter-controllers
+    connections between controllers in a way which maximizes the optical links connectivity
```
TheoLaudatQM, line 10: A . is missing after connectivity I guess

qualang_tools/simulator/simulator.py
```diff
@@ -0,0 +1,52 @@
...
+    n_controllers, actual_controllers=None, n_connections=12, print_debug=False
+):
+    """
+    Creates a list of :class:`~qm.simulate.interface.ControllerConnection` objects with the Inter-controllers
+    connections between controllers in a way which maximizes the optical links connectivity
+    For use inside the :class:`~qm.simulate.interface.SimulationConfig` for simulating the controllers` behavior.
+
+    :param n_controllers: The number of controllers
+    :type n_controllers: int
+    :param actual_controllers: A list of the controllers actually being used in the simulation
```
TheoLaudatQM, line 15: Maybe worth specifying that the index starts at 1

qualang_tools/simulator/simulator.py
```diff
@@ -0,0 +1,52 @@
...
+    :returns: The ControllerConnection object
+
+    """
+    if not actual_controllers:
+        actual_controllers = [i + 1 for i in range(n_controllers)]
+    controller_connections = []
+    unused_connection = np.ones((n_controllers, n_connections), dtype=bool)
+    while unused_connection.any():
+        for i in range(n_controllers):
+            for j in range(i + 1, n_controllers):
```
TheoLaudatQM, line 30: if n_controllers is 1 and actuall_controllers is not None as in `create_simulator_controller_connections(1, [1]),` then the script will hang forever. I know it is not meant to happen, but maybe raising an error/warning could avoid some customer issues :)

TheoLaudatQM (commented):
Looks good to me, just a few comments before merging

qualang_tools/simulator/README.md
yomach: nice catch!

qualang_tools/simulator/simulator.py
yomach: Made it to return an empty list, which is also the expected output

qualang_tools/simulator_tools.py
yomach: FutureWarning - this returns the warning to the user in the console
DeprecationWarning - this strikethrough the function name when it's in the IDE.

It's a bit annoying we need to use both, but it works

TheoLaudatQM (approved):

PR #185, opened 2024-01-26
Add sampling rate to the waveform tools

TheoLaudatQM:
The idea is to be able to use these waveform functions with the OPX1000 that has a sampling rate of 2GS/s

### DIFF
--- a/qualang_tools/config/waveform_tools.py
+++ b/qualang_tools/config/waveform_tools.py
@@ -140,9 +140,12 @@ def flattop_gaussian_waveform(amplitude, flat_length, rise_fall_length, return_p
     :param str return_part: When set to 'all', returns the complete waveform. Default is 'all'. When set to 'rise',
     returns only the rising part. When set to 'fall', returns only the falling part. This is useful for separating
     the three parts which allows scanning the duration of the flat part is to scanned from QUA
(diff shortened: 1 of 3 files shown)


qualang_tools/config/waveform_tools.py
```diff
@@ -138,14 +140,16 @@ def flattop_gaussian_waveform(amplitude, flat_length, rise_fall_length, return_p
...
-
-    rise_part = gauss_wave[:rise_fall_length]
+    gauss_wave = amplitude * gaussian(
+        int(np.round(2 * rise_fall_length * sampling_rate / 1e9)), rise_fall_length / 5 * sampling_rate / 1e9
+    )
+    rise_part = gauss_wave[: int(rise_fall_length * sampling_rate / 1e9)]
     rise_part = rise_part.tolist()
     if return_part == "all":
-        return rise_part + [amplitude] * flat_length + rise_part[::-1]
+        return rise_part + [amplitude] * int(flat_length * sampling_rate / 1e9) + rise_part[::-1]
```
yomach, line 152: Unlike the previous functions, this only supports sampling rate which are above 1e9 (and specifically, only integer multiples of 1e9).
Need to make this more explicit & document (or support more)

Same for all flattop functions

yomach (approved):

PR #186, opened 2024-02-02
Raise error when log step is too small for integers

TheoLaudatQM:


tests_against_server/test_loops.py
```diff
@@ -343,8 +349,34 @@ def prog_maker(logspace_param):
...
+
+    cfg = deepcopy(config)
+    arange_param_list = [
+        [np.log10(4), np.log10(10000), 60],
+    ]
+    for param in arange_param_list:
+        print(f"Test qua_list with {param}:")
+        try:
+            simulate_program_and_return(cfg, prog_maker(param))
+            raise Exception("Didn't catch error!")
```
deanpoulos, line 377: It looks like you're trying to assert that an error occurred due to a bad parameter. There is a neat way to do this using pytest. Let's say you expect a `ValueError`. You can run:

```
with pytest.raises(ValueError):
   simulate_program_and_return(cfg, prog_maker(param))
```

Not only is it neater, but it allows you to specify the exact error you expect. Often I've used this and it's been able to catch my code raising a _different_ error for some completely other reason than the expected one.

qualang_tools/loops/loops.py
```diff
@@ -90,12 +90,17 @@ def from_array(var, array):
                 "When using logarithmic increments with QUA integers, the resulting values will slightly differ from the ones in numpy.logspace() because of rounding errors. \n Please use the get_equivalent_log_array() function to get the exact values taken by the QUA variable and note that the number of points may also change."
             )
             if step > 1:
-                return (
-                    var,
-                    round(start),
-                    var < round(stop) * np.sqrt(float(step)),
-                    Cast.mul_int_by_fixed(var, float(step)),
-                )
+                if int(round(start) * float(step)) == int(round(start)):
```
deanpoulos, line 93: Should this be a mutiplication or addition?

In my head, you should check to see if the ith variable and i+1th value are the same, i.e.
`if int(round(start) + float(step)) == int(round(start)):`.

Also, this should be checked not just for the 0th and 1st, but all successive values?

Maybe I'm confused because logspace and linspace start/step represent different things...

qualang_tools/loops/loops.py
```diff
@@ -249,12 +254,17 @@ def qua_logspace(var, start, stop, num):
             "When using logarithmic increments with QUA integers, the resulting values will slightly differ from the ones in numpy.logspace() because of rounding errors. \n Please use the get_equivalent_log_array() function to get the exact values taken by the QUA variable and note that the number of points may also change."
         )
         if step > 1:
-            return (
-                var,
-                round(10**start),
-                var < round(10**stop * np.sqrt(step)),
-                Cast.mul_int_by_fixed(var, float(step)),
-            )
+            if int(round(10**start) * float(step)) == int(round(10**start)):
```
deanpoulos, line 257: Same here, just hit resolve if I'm wrong

tests_against_server/test_loops.py
```diff
@@ -95,20 +90,8 @@ def IQ_imbalance(g, phi):
     }
 
 
-def simulate_program_and_return(config, prog, duration=50000):
-
+def simulate_program_and_return(config, prog):
     qmm = QuantumMachinesManager()
```
deanpoulos, line 94: How can you run this without supplying any arguments to the constructor? Are you trying to use the conftest? If so, you can just put `qmm` as the first argument to the test function, i.e.,

`def simulate_program_and_return(**qmm,** config, prog):`

qualang_tools/loops/loops.py
TheoLaudatQM: With log scan the syntax is `a[i+1] = step * a[i]` so here I just want to check that the step is large enough so that two successive value are not equal, otherwise the program will hang (a[i+1] = a[i]). When step>1, checking the first iteration is enough because the difference between two successive values will increase

qualang_tools/loops/loops.py
TheoLaudatQM: Same as above :)

tests_against_server/test_loops.py
TheoLaudatQM: Very nice, I just didn't know about it :)
Feel free to modify it if it is not too complicated

tests_against_server/test_loops.py
```diff
@@ -197,6 +176,7 @@ def prog_maker(list_param):
 
     cfg = deepcopy(config)
     arange_param_list = [
```
deanpoulos, line 178: If you're trying to run the same test against multiple input values, you can use the `pytest.parameterize` decorator. This has the benefit of giving you better information if you ever start gathering statistics on your unit tests, and will also help you to find out which set of input parameters failed. It also allows you to not have to use a for-loop. It's not really essential here since these are server-tests, but it might be worth doing in the future if you don't already know about it

deanpoulos: Here is an example of the syntax in the pull request for `long_wait`:
```python
@pytest.mark.parametrize("wait_time", [4, 16, 100,
                                       dummy_max_wait_time-1,
                                       dummy_max_wait_time,
                                       dummy_max_wait_time+1,
                                       100*dummy_max_wait_time])
def test_long_wait_time_simulation(qmm, config, wait_time):
    """ Extracts the wait time from simulation and asserts it to be equal to the input wait time."""
    op = "playOp"

    with program() as prog:
        play(op, "qe1")
        long_wait(wait_time, threshold_for_looping=dummy_max_wait_time)
    ...
```

tests_against_server/test_loops.py
```diff
@@ -228,27 +208,61 @@ def prog_maker(list_param):
...
+
+    cfg = deepcopy(config)
+    arange_param_list = [
+        [np.logspace(np.log10(4), np.log10(10000), 60), "int"],
+    ]
+    for param in arange_param_list:
+        print(f"Test qua_list with {param}:")
+        try:
+            simulate_program_and_return(cfg, prog_maker(param))
+            raise Exception("Didn't catch error!")
```
deanpoulos, line 252: Could also use `pytest.raises` here too

qualang_tools/loops/loops.py
deanpoulos: Oh yep I see that now. All good.

tests_against_server/test_loops.py
deanpoulos: Ok that's fine. But on line 94, you leave the inputs empty. When I run that on my machine I get this:

```python
>>> from qm import QuantumMachinesManager
2024-04-10 22:24:00,708 - qm - INFO     - Starting session: 7c33597d-ada4-4350-8bd5-35ff41f888b1
>>> qmm = QuantumMachinesManager()
2024-04-10 22:24:06,152 - qm - ERROR    - Failed to detect a QuantumMachines server. Tried connecting to :80.
...
```

tests_against_server/test_loops.py
deanpoulos: It's fine for now since it works for whoever tested. it. But at least you can use it in the future

TheoLaudatQM: I updated it just to get the gist of how it works, thanks!

tests_against_server/test_loops.py
TheoLaudatQM: Yes I leave it empty once I am done so that the whole world doesn't see our IP address and port

tests_against_server/test_loops.py
TheoLaudatQM: Updated, thanks!

TheoLaudatQM:
Is it ready to be merged @deanpoulos or do you have other comments?

deanpoulos (approved):

PR #189, opened 2024-03-04
Fix the units of phase and add a flag to allow wrapped phase

SamarthH:
- Fixed wrong unit of phase in the driver (units were initially rad, but because they are being converted using a 180/pi, units should be deg)
- Added a flag ``unwrap_phase'' to allow selection of whether or not to allow phase to be unwrapped before returning


yomach (approved):
Looks good, thanks!

TheoLaudatQM (approved):
All good, thanks for the modification!

PR #191, opened 2024-03-26
Improve wait_until_job_is_paused

TheoLaudatQM:
Add warning when timeout is reached

### DIFF
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -121,7 +121,7 @@ def progress_counter(iteration, total, progress_bar=True, percent=True, start_ti
         print("")
 
 
-def wait_until_job_is_paused(running_job: RunningQmJob, timeout: int = 30):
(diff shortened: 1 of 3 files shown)


qualang_tools/results/results.py
```diff
@@ -136,4 +137,6 @@ def wait_until_job_is_paused(running_job: RunningQmJob, timeout: int = 30):
     while (not running_job.is_paused()) and (delay < timeout):
         time.sleep(0.1)
         delay = time.time() - start
+    if delay < timeout:
```
yomach, line 140: shouldn't it be `delay > timeout`?
Also, we might want it to be an exception. Or at least, add a `strict_timeout` boolean argument, which if set to `true`, throws an exception.
WDYT?

TheoLaudatQM: Oh wow very good point I was too tired I guess...
Yeah you're probably right, I wanted to minimize the arguments but throwing an exception by default is probably more robust and closer to what customers will want

yomach (approved):

PR #197, opened 2024-04-17
Fixed printing bug ParameterTable

arthurostrauss:
Fixed bug emerging wrong usage of the __repr__ class method


TheoLaudatQM (approved):
This is good :)

PR #199, opened 2024-04-19
fix bug in Datahandler

TheoLaudatQM:
data_path = Path(self.root_data_folder) / node_contents["metadata"]["data_path"]


nulinspiratie (approved):
Thanks @TheoLaudatQM , I also went through the two files and couldn't find any other problematic occurrences, so should be good

PR #200, opened 2024-04-19
fix: only add xarray data processor if installed

nulinspiratie:
The Xarray data processor was added as a default data processor for the data handler. This raises an error if a save is performed and xarray isn't installed.

The data processor is now only added by default if xarray is installed

### DIFF
--- a/qualang_tools/results/data_handler/data_processors.py
+++ b/qualang_tools/results/data_handler/data_processors.py
@@ -187,7 +187,7 @@ class XarraySaver(DataProcessor):
 
 
 try:
-    import xarray
+    import xarray  # noqa: F401
 
     DEFAULT_DATA_PROCESSORS.append(XarraySaver)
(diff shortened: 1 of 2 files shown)


nulinspiratie:
@TheoLaudatQM could you test if this fixes the problem?

TheoLaudatQM (approved):
Tested and working well, thanks!

qualang_tools/results/data_handler/data_processors.py
```diff
@@ -186,4 +186,9 @@ def post_process(self, data_folder: Path):
                 array.to_netcdf(data_folder / path.with_suffix(self.file_suffix))
 
 
-DEFAULT_DATA_PROCESSORS.append(XarraySaver)
+try:
```
yomach, line 189: I don't think this is the right solution... I believe this part should be in the init file

PR #205, opened 2024-04-26
samplling rate for integration weights

KevinAVR:
I created this PR to upgrade the integration weights tool to receive a sampling rate, from the optimal weight code when it takes every 4 ns it enforces that I sample every clock cycle with sliced demod -- now I can have the freedom -- I made it backward comptible -- if you approve I can add the chnge log

### DIFF
--- a/qualang_tools/config/integration_weights_tools.py
+++ b/qualang_tools/config/integration_weights_tools.py
@@ -24,6 +24,28 @@ def convert_integration_weights(integration_weights, N=100, accuracy=2**-15, plo
     :type plot: bool
     :return: List of tuples representing the integration weights
     """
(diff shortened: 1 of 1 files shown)


yomach (changes requested):
It's hard for me to understand exactly what you did here, and I'm not sure it would work (and what is the usecases).
Can you please add the changelog (that would make it easier to understand the reason) and also some tests?

As for the syntax, there are nicers ways to do it, I don't like adding a new function just because we've added another parameter...
What would be next `convert_integration_weights_with_sample_time_and_a_new_parameter` ?

I'll show an alternative to this next week, if it's not urgent...

PR #206, opened 2024-04-28
Bugfix/long wait units import

deanpoulos:
Enforce `threshold_for_looping` to be an integer so that the final wait time is also an integer.


yomach (commented):
I'm not a fan of this, in QUA we automatically convert 1e4 (which is a float) to an int and use it. I don't like it that qualang_tools behaves differently...

With that being said, I'm not completely opposed to it. So what is the reasoning here?

deanpoulos:
> I'm not a fan of this, in QUA we automatically convert 1e4 (which is a float) to an int and use it. I don't like it that qualang_tools behaves differently...
> 
> With that being said, I'm not completely opposed to it. So what is the reasoning here?

So the scope of this bugfix is limited to the new `long_wait` macro that I wrote. Despite my testing, when I tried to run it, the QUA compiler complained that my wait_time was _not_ an integer.

This is _despite_ me using `unit.(coerce_to_integer=True).to_clock_cycles(1e6)`, which is unusual, so I decided to go further and write `unit.(coerce_to_integer=True).to_clock_cycles(1_000_000)`, as well as throw an error if the `threshold_for_looping` parameter is a float.

yomach:
> > I'm not a fan of this, in QUA we automatically convert 1e4 (which is a float) to an int and use it. I don't like it that qualang_tools behaves differently...
> > With that being said, I'm not completely opposed to it. So what is the reasoning here?
> 
> So the scope of this bugfix is limited to the new `long_wait` macro that I wrote. Despite my testing, when I tried to run it, the QUA compiler complained that my wait_time was _not_ an integer.
> 
> This is _despite_ me using `unit.(coerce_to_integer=True).to_clock_cycles(1e6)`, which is unusual, so I decided to go further and write `unit.(coerce_to_integer=True).to_clock_cycles(1_000_000)`, as well as throw an error if the `threshold_for_looping` parameter is a float.

Please write a test (put it in the test folder) that recreates this issue.
There might be a bug in the `unit` class, and there might also be a bug in the OPX.

TheoLaudatQM:
There was indeed a bug with th u.to_clock_cycles function that doesn't cast 1e3//4 to an integer...
I create a separated PR to fix it: [https://github.com/qua-platform/py-qua-tools/pull/214](https://github.com/qua-platform/py-qua-tools/pull/214l)

TheoLaudatQM (approved):

PR #211, opened 2024-05-08
control_panel - Fix rounding error that caused voltage drifts

matpompili:
A rounding error in the voltage book-keeping causes voltage drifts when switching in between two voltages. An example program switches between -0.05V and 0.05V (factor of two in the scope is because of missing 50Ω termination).

Essentially, every `qua.play` instruction is missing one unit of DAC voltage (2^-16), so on repeated calls the voltage drifts.
This PR fixes the error. It also replaces `np.round` with `round` as it is a factor of ~30 faster.

Before the fix:
image

After the fix:
image

Remember to:
* Update the CHANGELOG.md
* Added tests for the feature or fix


yomach:
Hi @matpompili,
Thanks!
I'm slightly confused by this. Following the math, I'm failing to see the issue.
If I do:
`round(-0.05 / 2**-16) * 2**-16 + round(0.05 / 2**-16) * 2**-16`, which is applying your voltages with the current rounding, then I get 0.
If I apply your fix, I get something else (two LSB units, 2*2**-16)
I will check on our simulator

yomach:
Hi @matpompili 
After some debugging, the issue was not taking into effect the rounding (down) that is happening on the Pulse Processor. I've pushed a commit to your branch that fixes it. It seems to solve the issue.

Your fix, while worked in this case, might produce drift in other cases.

Can you please test it as well?

matpompili:
Thanks @yomach it works!

I made a couple of small edits:
- Removed the if statement by making two functions
- Converted `np` calls to `math` since they are faster on scalars.

Tested on setup, good to merge from our end!

yomach (approved):

PR #216, opened 2024-05-18
Refactor: Move DEFAULT_DATA_PROCESSORS

nulinspiratie:
Move `DEFAULT_DATA_PROCESSORS` from `data_processors.py` to `__init__.py`,

In response to comment by @yomach

### DIFF
--- a/qualang_tools/results/data_handler/__init__.py
+++ b/qualang_tools/results/data_handler/__init__.py
@@ -1,6 +1,23 @@
+from abc import ABC
+
 from .data_folder_tools import *
 from . import data_processors
-from .data_processors import DEFAULT_DATA_PROCESSORS
+
 from .data_handler import *
+
+
+DEFAULT_DATA_PROCESSORS = [
+    data_processors.MatplotlibPlotSaver,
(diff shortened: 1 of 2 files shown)


nulinspiratie:
Actually, I can't get the tests working. The problem is the following:
- `results/data_handler/data_handler.py` needs `DEFAULT_DATA_PROCESSORS` for one of its class attributes.
- `DEFAULT_DATA_PROCESSORS` is defined in `results/data_handler/__init__.py`
- `results/data_handler/__init__.py` loads `results/data_handler/data_handler.py` 
This leads to a circular import.

I can't see a way around it. @yomach if you can't either then I suggest to cancel this PR, and keep `DEFAULT_DATA_PROCESSORS` in `data_processors.py`

yomach:
@nulinspiratie See my changes.
Also note, I've put all of the `__all__` under the `__init__.py`, as it is more customary to do

deanpoulos (dismissed):

MichalGQM:
@nulinspiratie is it done? can you resolve and merge it?

PR #217, opened 2024-05-23
add import numpy as np

TheoLaudatQM:


yomach (commented):
Add to changelog please.
QCoDeS - Fixed compatibility with qm-qua 1.1.6 or newer

TheoLaudatQM:
It is actually just in the examples folder, but I added it to the changelog

yomach (approved):

yomach:
Oh, in this case it's not really needed, but ok.
I missed that, sorry

PR #220, opened 2024-06-05
Fix voltage book-keeping error introduced in #211

matpompili:
In #211 I introduced an error: when calling set_amplitude with value 0.0, the value on the OPX is correctly updated, but the record in the python program is not, breaking following updates.

Remember to:
- [x] Update the CHANGELOG.md
- ~~Added tests for the feature or fix~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Fixed
+- control_panel - Fix voltage booking error introduced in the previous fix.
 - control_panel - Fix rounding error in `ManualOutputControl` that caused voltage drifts.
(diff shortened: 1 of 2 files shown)


yomach:
Hi @matpompili,
Thanks for the PR and for fixing the bug.
While I originally wrote this, I think it is confusing to use only the `value` variable and change it's meaning. I prefer the creation of more variables such that the code behavior is explicit.

Please see my changes and let me know what you think.

matpompili:
@yomach I think your version doesn't work: the qua program behaves differently depending on the value:
- If the value is 0.0, it ramps the voltage back to zero.
- If the value is not 0.0, it adds that value to the current voltage.

Your version always sends the set_value (the actual voltage) but that's not what the qua program expects.
I agree that it is a bit confusing.

yomach:
Yes, you're right...
I'll fix it

matpompili:
LGTM, will test on hardware and let you know.

matpompili:
Tested, works as expected for positive, negative and zero voltages.

yomach (approved):

PR #221, opened 2024-06-24
Copy-paste Theo's virtual_gate_sequence class into qualang-tools.

deanpoulos:
Introduces a class for designing sequences of pulses in voltage-space by allowing the definition of many sets of different voltage levels for many channels.

This was developed by @TheoLaudatQM in qua-libs, and has been copy-pasted here in preparation for removing the duplication that exists in qua-libs.

Since then, I have renamed it with Theo's approval to `VoltageGateSequence` and added a helpful docstring.

It has been well-tested in a few labs, but requires some auto-testing for edge cases which involve large jumps and short gates.


qualang_tools/virtual_gates/virtual_gate_sequence.py
```diff
@@ -0,0 +1,252 @@
...
+class VoltageGateSequence:
+    def __init__(self, configuration: Dict, elements: List[str]):
+        """
+        Initializes a VirtualGateSequence object for designing arbitrary pulse sequences using virtual gates.
+
+        This class allows the creation of complex pulse sequences by defining voltage levels (points) in gate-space
+        and durations for multiple elements. It supports the addition of steps and ramps, and keeps track of average
+        voltage levels for compensation pulses.
+
+        The `configuration` provided will be updated to include necessary operations and waveforms for the sequence.
```
yomach, line 17: Which means that this class must be called before opening the QM, right? Should be mentioned (probably in the readme file).
Or even more, that the program itself should be written before opening the qm

qualang_tools/virtual_gates/virtual_gate_sequence.py
```diff
@@ -0,0 +1,252 @@
...
+
+                if not self.is_QUA(ramp_duration):
+                    ramp_rate = 1 / ramp_duration
+                    play(ramp((voltage_level - self.current_level[i]) * ramp_rate), gate, duration=ramp_duration >> 2)
+                    wait(_duration >> 2, gate)
+
+            self.current_level[i] = voltage_level
+
+    def add_compensation_pulse(self, duration: int) -> None:
+        """Add a compensation pulse of the specified duration whose amplitude is derived from the previous operations.
```
yomach, line 207: I guess that this would only work when everything is in strict timing?

yomach (commented):
Need to add `readme.md` for this function, as well as in the main one.

qualang_tools/virtual_gates/virtual_gate_sequence.py
deanpoulos: Added a note in README please check

qualang_tools/virtual_gates/virtual_gate_sequence.py
deanpoulos: Added a note in the README, please check

qualang_tools/voltage_gates/README.md
```diff
@@ -0,0 +1,202 @@
...
+> **Note**: Make sure to initialize the class before opening the `qm`!
+
+> **Note**: Make sure the check the timing of the pulses in simulation or use `strict_timing()` to avoid real-time gaps!
+
+## Features
+
+- **Dynamic Voltage Control**: Define voltage levels (points) in gate-space and durations for multiple elements.
+- **Step and Ramp Support**: Add steps and ramps to the pulse sequence.
+- **Compensation Pulse**: Keeps track of average voltage levels for compensation pulses.
+- **Configuration Management**: Updates the provided configuration with necessary operations and waveforms for the sequence.
```
TheoLaudatQM, line 14: I am not sure, but maybe write Keep and Update to be compatible with the infinitive of the bullet points above.

qualang_tools/voltage_gates/README.md
```diff
@@ -0,0 +1,202 @@
...
+# List of elements involved in the gate operations
+elements = ["P5_sticky", "P6_sticky"]
+
+# Initialize the VoltageGateSequence
+seq = VoltageGateSequence(configuration, elements)
+```
+
+## Methods
+
+### `add_step`
```
TheoLaudatQM, line 39: I would show the `add_points` function before the `add_step` to illustrate what voltage_point_name means and how to set it up. I think that adding a simple charge stability map picture with relevant points would be helpful to illustrate the purpose, what do you think?

qualang_tools/voltage_gates/README.md
```diff
@@ -0,0 +1,202 @@
...
+    self,
+    level: list[Union[int, QuaExpression, QuaVariable]] = None,
+    duration: Union[int, QuaExpression, QuaVariable] = None,
+    voltage_point_name: str = None,
+    ramp_duration: Union[int, QuaExpression, QuaVariable] = None
+) -> None:
+```
+
+- **level**: Desired voltage level of the different gates composing the gate in Volts.
+- **duration**: How long the voltage level should be maintained in ns. Must be a multiple of 4ns and larger than 16ns.
```
TheoLaudatQM, line 54: These can also be QUA variables

qualang_tools/voltage_gates/README.md
```diff
@@ -0,0 +1,202 @@
...
+def add_compensation_pulse(self, duration: int) -> None:
+```
+
+- **duration**: Duration of the compensation pulse in clock cycles (4ns). Must be larger than 4 clock cycles.
+
+**Example Usage**:
+
+```python
+# Add a compensation pulse with a duration of 1000 ns
+seq.add_compensation_pulse(duration=1000)
```
TheoLaudatQM, line 90: I think that this can be improved by setting the amplitude of the pulse instead and have the duration derived behind the scene (with slight adjustment of the amplitude for the duration to be a multiple of 4ns). Otherwise you can get an error because the needed amplitude is larger than 0.5V for the  specified duration...

TheoLaudatQM (commented):
Very nice!
Just a few comments and the compensation pulse improvement could also be done in a second step if needed

qualang_tools/voltage_gates/README.md
deanpoulos: Done

qualang_tools/voltage_gates/README.md
deanpoulos: Yes, good idea

deanpoulos: Done

qualang_tools/voltage_gates/README.md
deanpoulos: Noted

qualang_tools/voltage_gates/README.md
deanpoulos: Are you proposing a change to the code or the README? I don't see how it's possible with the current API but I might be missing something.

deanpoulos: @TheoLaudatQM

TheoLaudatQM: Oh sorry I missed this...
I meant in the code yes, I think it is just about modifying the function `add_compensation_pulse(self, duration: int)` so that the duration is inferred from the amplitude and not the opposite.
This being said it will require some time to test it properly so I suggest to add this functionality at a later stage in a new PR, what do you think?

deanpoulos: No problem. I think you're right that it's best if the duration is inferred from the amplitude, and not the opposite.

Let's leave it for future work so that we can merge this and integrate properly to qua-libs.

deanpoulos: @TheoLaudatQM could you please approve?

TheoLaudatQM (approved):
Sorry for the delay :)

PR #223, opened 2024-06-27
Fix/remove qm-qua 1.2.0 deprecations

deanpoulos:
Fixed deprecated imports:

- `from qm.QuantumMachinesManager import QuantumMachinesManager` -->
`from qm import QuantumMachinesManager`
- `from qm.QuantumMachine import QuantumMachine` --> `from qm import QuantumMachine`
- `from qm.QmJob import QmJob` --> `from qm import QmJob`


deanpoulos:
@yonatanrqm could you please review/approve this today? Should take 30s

yomach (approved):
I suggest we look at all of the deprecation, even those not removed soon, and fix those as well

deanpoulos:
> I suggest we look at all of the deprecation, even those not removed soon, and fix those as well

I looked into them and made a list. I'll make a PR for it at a late date.

PR #226, opened 2024-08-08
Create the root folder if it doesn't exist

TheoLaudatQM:


deanpoulos (approved):

nulinspiratie (changes requested):
I like it but it would be good to print a message if the folders need to be created

nulinspiratie (approved):

PR #235, opened 2024-10-23
Bump version to 0.17.8

deanpoulos:
# [0.17.8] - 2024-10-23
## Added
- results - Allow the data saver to create the root folder if it doesn't exist.
- wirer - Automatic tool to allocate channels for arbitrary combinations of QM instruments and superconducting qubits.
## Fixed
- data_handler - Fix figure saving cutting off title text if it is long using `bbox_inches="tight"`.


deanpoulos:
Closing in favor of incrementing the minor version instead (0.18.0 instead of 0.17.8), see #236

PR #237, opened 2024-10-23
Fix/refactor videomode folder

TheoLaudatQM:
Rename the module called video_mode into live_mode to avoid confusion with the video mode tool used to tune up quantum dots devices.
We can still debate on the name if you think that another one would be more relevant for this tool which is used for updating parameters from a QUA program while the program is running in an asynchronous manner.

- [ ] Deprecation warnings when trying to import qualang_tools.video_mode

### DIFF
--- a/tests/test_livemode.py
+++ b/tests/test_livemode.py
@@ -1,8 +1,8 @@
-# Tests for the video mode functionality
+# Tests for the live mode functionality
 import pytest
 from qm.qua import *
 import numpy as np
-from qualang_tools.video_mode import ParameterTable
+from qualang_tools.live_mode import ParameterTable
 
 
 def gauss(amplitude, mu, sigma, length):


nulinspiratie (commented):
What about if we have the `video_mode/__init__.py` import the original functions/classes which are now in live_mode? This way the code remains backwards compatible

TheoLaudatQM:
@nulinspiratie What I had in mind was this
```
import warnings
warnings.warn("the video_mode module is deprecated, please import live_mode instead", DeprecationWarning,
              stacklevel=0)
from qualang_tools.live_mode.livemode import LiveMode, ParameterTable

__all__ = ["LiveMode", "ParameterTable"]
```
But 

1. the warning doesn't show up when calling `from qualang_tools.video_mode import *`
2. The user will still get error when running his previous scripts that most likely use `from qualang_tools.video_mode import VideoMode, ParameterTable`

nulinspiratie:
@TheoLaudatQM looks like a nice solution! All good from my end

nulinspiratie (approved):

yomach:
> @nulinspiratie What I had in mind was this
> 
> ```
> import warnings
> warnings.warn("the video_mode module is deprecated, please import live_mode instead", DeprecationWarning,
>               stacklevel=0)
> from qualang_tools.live_mode.livemode import LiveMode, ParameterTable
> 
> __all__ = ["LiveMode", "ParameterTable"]
> ```
> 
> But
> 
>     1. the warning doesn't show up when calling `from qualang_tools.video_mode import *`
> 
>     2. The user will still get error when running his previous scripts that most likely use `from qualang_tools.video_mode import VideoMode, ParameterTable`

Not sure about 1, I am 99% it is possible to display a warning in this case.
This can be solved by doing:
`from qualang_tools.live_mode.livemode import LiveMode as VideoMode`.

We can merge if we're okay with the breaking changes and the lack of backward compatibility. Otherwise, we can make it compatible.

nulinspiratie:
Hmm, if we want to maintain backwards compatibility, we could also opt for the following:

**video_mode/videomode.py**
```python
from qualang_tools.live_mode import LiveMode
import warnings


class VideoMode(LiveMode):
    def __init__(self, qm: QuantumMachine, parameters: Union[Dict, ParameterTable]):
        warnings.warn("the video_mode module is deprecated, please import live_mode instead", DeprecationWarning,
              stacklevel=0)
        super().__init__(self, qm, parameters)
```

We would then need to rename the new `VideoMode` to something else, maybe `VideoMode2D`?

PR #239, opened 2024-10-24
Feature/2q rb

deanpoulos:
# Overview
Migrated 2Q RB from qua-libs, to go under `qualang_tools/characterization/two_qubit_rb` adding to the README and CHANGELOG, and fixing formatting. I added the module `characterization` module in preparation for XEB which might come soon too.

# Cirq Dependency ~~Issue~~
~~There is an issue with adding cirq as a dependency, since cirq requires python>=3.10, but qualang_tools supports >=3.8.~~

~~I added `two_qubit_rb` as an optional dependency, i.e.,  
```pip install qualang_tools[two_qubit_rb], ```~~

~~but this doesn't help the situation.~~

~~So, the cirq import is always performed with a graceful ImportError indicating the correct version to install.~~

`cirq` is now installed with the `two-qubit-rb` extra when the python version is >3.10, as guaranteed by the following lines in the pyproject.toml:
```toml
two-qubit-rb = ["xarray", "cirq", "tqdm"]

[tool.poetry.dependencies.cirq]
version = "1.2.0"
python = ">=3.10"
```

I verified that `cirq` will be installed when running:
```bash
poetry install -E two-qubit-rb
```

~~I have a regression test which creates a TwoQubitRb instance and uses it to generate some sequences, verifying that they are valid. But, this will not be auto-tested since `cirq` is a required dependency, and therefore I'm forced to skip this test.~~


yomach:
Please list again cirq as a dependency, I want to see what is failing, I am positive we can fix it.

qualang_tools/characterization/two_qubit_rb/README.md
```diff
@@ -0,0 +1,192 @@
...
...t Clifford operations (576), the CNOT-like class (5184), the iSWAP-like class (5184) and the SWAP-like class (576). In this use-case example we introduce an implementation on the OPX+ {'client': '1.1.3', 'server': '2.60-5ba458f'} using the current version (2023, July) of the generic `TwoQubitRb` class. The implementation exploits the [baking](https://github.com/qua-platform/py-qua-tools/blob/main/qualang_tools/bakery/README.md) tool to generate the individual Clifford operations. The class then uses the [Input Stream](https://docs.quantum-machines.co/0.1/qm-qua-sdk/docs/Guides/features/?h=declare_input_stream#input-streams) feature to send a string of Clifford indices to the OPX that represent the executed gate sequence which is terminated with the inverse operation. The execution is based on the [Switch Case](https://docs.quantum-machines.co/0.1/qm-qua-sdk/docs/Guides/features/?h=switch#switch-case) flow control of QUA, which sets the current minimal gate duration limit to 40 ns.  <!--The inverse is calculated in Python using Clifford tableaus. An updated version of the TwoQubitRb class can be found in the [py-qua-tools](https://github.com/qua-platform/py-qua-tools) repository.-->
```
yomach, line 3: Please remove this: `OPX+ {'client': '1.1.3', 'server': '2.60-5ba458f'}`
1. At a minimum, just write `qm-qua=1.1.3, QOP=2.2.0`
2. Since this is no longer a usecase, I would remove it and make sure it works with the latest qm-qua and QOP (OPX+ & OPX1000), and if it doesn't, note it.

yomach: Link to input stream (and other docs) needs to be updated with `latest` instead of `0.1`.

yomach: Last sentence needs to be removed

qualang_tools/characterization/two_qubit_rb/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+```python
+res.plot_fidelity()
+```
+
+### Under the Hood: Clifford Sequence Generation
+#### How are all the 11,520 2Q Cliffords loaded onto the OPX?
+In order to both:
+ 1. Efficiently generate random two-qubit clifford sequences and the recovery gate
+ 2. Use minimal OPX resources within the compiled program,
```
yomach, line 115: `,` at the end?

qualang_tools/characterization/two_qubit_rb/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### How do commands combine to make Cliffords?
+Each two-qubit Clifford can be decomposed into 1/720 symplectic gates (commands 0-719), followed by 1/16 Pauli gates (commands 720-735).
+
+#### Why use commands instead of baking all the Cliffords?
+It would cost too much waveform memory to pre-upload all 11,520 two-qubit Clifford gates onto the OPX. Ideally, we could just bake all possible single-qubit and two-qubit gates, and this would be the most memory-friendly and efficient way to reconstruct 2Q Cliffords on the fly. However, the real-time switch-case on the OPX has a certain latency, which can be greater than the duration of a single gate, leading to gaps between consecutive gates. Therefore, the extended length of a "command" and smaller volume of commands compared with the two-qubit Cliffords provide an efficient middle-ground for navigating the OPX's unique constraints.
```
yomach, line 130: Now that it is 16ns, is it still true?
Well obviously, in some cases people go lower, but this is not common...

In any case, please edit it to make it more friendly to the OPX+ :)

README.md
```diff
@@ -68,6 +68,13 @@ Note that in order use the `Interactive Plot Library`, you need to install using
 pip install --upgrade qualang-tools[interplot]
 ```
 
+Note that in order to run 2-Qubit Randomized Benchmarking, you need to install using
+```bash
+pip install --upgrade qualang-tools[two-qubit-rb]
+# Install cirq separately, since it only works with python==3.10 and can't be listed as a dependency
```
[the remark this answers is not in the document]
deanpoulos: I managed to fix it, but as mentioned in the updated PR description, `cirq` will only install as a dependency when python>=3.10. Otherwise, it will be ignored and have to be pip installed manually. Do you think this is sufficient?

README.md
```diff
@@ -68,6 +68,13 @@ Note that in order use the `Interactive Plot Library`, you need to install using
 pip install --upgrade qualang-tools[interplot]
```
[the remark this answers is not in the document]
deanpoulos: Added the Data Handler, which was missing, and added a node that the config builder is deprecated. You said many are missing- did I miss any?

README.md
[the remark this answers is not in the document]
deanpoulos: Actually, `poetry install` is only working when in Python >3.10, it runs into some other issue on Python 3.8. looking into it.

README.md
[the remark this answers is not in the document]
yomach: Good enough for now, thanks!
(Not sure if any others are missing)

yomach (approved):

PR #240, opened 2024-10-29
Feature: `Wirer` support fixed-freq transmons

deanpoulos:
## Added:
- CROSS_DRIVE (xd) lines
- ZZ_DRIVE (zz) lines
- test case

## Fixed:
- Bug where "untyped" lines would be added in non-deterministic order
- Bug where allocations on the same channel with different line types plotted overlapping

## Example
```python
    qubits = [1, 2]
    qubit_pairs = [(1, 2), (2, 1)]

    connectivity = Connectivity()

    connectivity.add_resonator_line(qubits=qubits)
    allocate_wiring(connectivity, instruments_2lf_2mw)

    connectivity.add_qubit_drive_lines(qubits=qubits)
    allocate_wiring(connectivity, instruments_2lf_2mw, block_used_channels=False)

    connectivity.add_qubit_pair_zz_drive_lines(qubit_pairs)
    allocate_wiring(connectivity, instruments_2lf_2mw, block_used_channels=False)

    connectivity.add_qubit_pair_cross_drive_lines(qubit_pairs)
    allocate_wiring(connectivity, instruments_2lf_2mw)

    visualize(connectivity.elements, instruments_2lf_2mw.available_channels)
```

image

### DIFF
--- a/qualang_tools/wirer/connectivity/connectivity_base.py
+++ b/qualang_tools/wirer/connectivity/connectivity_base.py
@@ -71,19 +71,6 @@ class ConnectivityBase:
 
         return elements
 
-    def _make_qubit_pair_element_pair(self, qubit_pairs: QubitPairsType):
-        if not isinstance(qubit_pairs, list):
-            qubit_pairs = [qubit_pairs]
-
-        elements = []
-        for qubit_pair in qubit_pairs:
-            id = QubitPairReference(*qubit_pair)
-            if id not in self.elements:
-                self.elements[id] = Element(id)
-            elements.append(self.elements[id])
-
-        return elements
-
     def _add_elements(self, elements: List[Element]):
         for element in elements:
             self._add_element(element)
--- a/qualang_tools/wirer/connectivity/connectivity_base.py
+++ b/qualang_tools/wirer/connectivity/connectivity_base.py
@@ -71,6 +71,19 @@ class ConnectivityBase:
 
         return elements
 
+    def _make_qubit_pair_element_pair(self, qubit_pairs: QubitPairsType):
+        if not isinstance(qubit_pairs, list):
+            qubit_pairs = [qubit_pairs]
+
+        elements = []
+        for qubit_pair in qubit_pairs:
+            id = QubitPairReference(*qubit_pair)
+            if id not in self.elements:
+                self.elements[id] = Element(id)
+            elements.append(self.elements[id])
+
+        return elements
+
     def _add_elements(self, elements: List[Element]):
         for element in elements:
             self._add_element(element)
--- a/qualang_tools/wirer/connectivity/wiring_spec.py
+++ b/qualang_tools/wirer/connectivity/wiring_spec.py
@@ -35,6 +35,7 @@ class WiringLineType(Enum):
     CHARGE = "q"
     COUPLER = "c"
     CROSS_DRIVE = "xd"
+    ZZ_DRIVE = "zz"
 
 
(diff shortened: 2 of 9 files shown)


qualang_tools/wirer/connectivity/wiring_spec.py
```diff
@@ -34,13 +34,17 @@ class WiringLineType(Enum):
     FLUX = "z"
     CHARGE = "q"
     COUPLER = "c"
+    CROSS_DRIVE = "xd"
```
HiroQM, line 37: This could be simply `CROSS_RESONANCE = "cr"` as we are adding `ZZ_DRIVE` as well? (I called it cross_drive to keep it applicable both to cross resonance and zz drive). I think I'm wondering if adding new enums as needed in the future would be what it is meant here.

qualang_tools/wirer/connectivity/wiring_spec.py
```diff
@@ -34,13 +34,17 @@ class WiringLineType(Enum):
...
+    ZZ_DRIVE = "zz"
 
 
 RESONATOR = WiringLineType.RESONATOR
 DRIVE = WiringLineType.DRIVE
 FLUX = WiringLineType.FLUX
 CHARGE = WiringLineType.CHARGE
 COUPLER = WiringLineType.COUPLER
+CROSS_DRIVE = WiringLineType.CROSS_DRIVE
+ZZ_DRIVE = WiringLineType.ZZ_DRIVE
```
HiroQM, line 47: (likewise)

qualang_tools/wirer/connectivity/connectivity_base.py
```diff
@@ -71,6 +71,19 @@ def _make_qubit_pair_elements(self, qubit_pairs: QubitPairsType):
...
+
+        elements = []
+        for qubit_pair in qubit_pairs:
+            id = QubitPairReference(*qubit_pair)
+            if id not in self.elements:
+                self.elements[id] = Element(id)
+            elements.append(self.elements[id])
+
+        return elements
+
```
HiroQM, line 86: Sorry, please remove this. (this comes from before the discussion where I thought we need to add two elements under qubit_pair to drive control and target. we will drive only control based on the discussion).

qualang_tools/wirer/wirer/wirer.py
```diff
@@ -39,20 +39,24 @@ def allocate_wiring(
         WiringLineType.FLUX,
         WiringLineType.CHARGE,
         WiringLineType.COUPLER,
+        WiringLineType.CROSS_DRIVE,
+        WiringLineType.ZZ_DRIVE,
```
HiroQM, line 43: (likewise)

HiroQM (commented):
I added comments to correct what I added at the beginning. If you could consider these. Thanks.

qualang_tools/wirer/connectivity/wiring_spec.py
deanpoulos: Thanks, "cr" is much cleaner than "xd".

Yes, for now I think adding new enums for new drives is useful as it gives then a standard wiring name, e.g. `q1-2.cr`, allowing it to be loaded and referenced later by QuAM in a standard way.

qualang_tools/wirer/connectivity/connectivity_base.py
deanpoulos: No problem. I removed that method that you added

HiroQM:
LGTM

HiroQM (approved):

PR #243, opened 2024-11-06
Fix/voltagecontrol compensation

TheoLaudatQM:
Fix the bug related to casting to integer the product V*T for the derivation of the compensation pulse


KevinAVR (approved):
I used in equal1 did not see in simulation that it workde but the values of 1000 and / 1000 are in the correct places :)

yomach:
If we want to be more efficient, we can bitshift by 10 instead of multiplying by 1000. It'll take 1 cycle instead of 6.

TheoLaudatQM:
Good point @yomach, I just updated it, thanks!

PR #244, opened 2024-11-06
Feat/voltagegates ramprate as a QUA variable

TheoLaudatQM:
Upon IBM's request, I adapted the class to allow the users to sweep the ramp rate in real time. I did basic testing using the simulator and everything seems to work and I added a warning, because the duration of the level after the ramp must be larger than 52ns because of a gap...


deanpoulos (approved):
Looks good. Is it worth adding an example usage + plot to the existing README under voltage_gates?

TheoLaudatQM:
Thanks @deanpoulos I didn't add an example or even a note on the readme because I didn't want people to start using it too much because of the gaps and synchronization issues it may cause...

deanpoulos:
@TheoLaudatQM will you merge this? We want to release a new py-qua-tools version today.

PR #248, opened 2024-11-20
Feat: Add unit.volts2raw

nulinspiratie:
This PR adds the method `unit.volts2raw` which converts voltage to raw data.
This is needed by QUAM for the implementation of time tagging.

I set that it should return float, is this correct or should it be an int?
If it's an int, would it make sense to either round it or cast to int?

### DIFF
--- a/qualang_tools/units/units.py
+++ b/qualang_tools/units/units.py
@@ -228,6 +228,14 @@ class unit:
         """
         return data * self.V / 4096
 
+    def volts2raw(self, Vp: Union[float, ndarray]) -> Union[float, ndarray]:
+        """Converts the volts to raw data.
+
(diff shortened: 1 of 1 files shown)


yomach:
`raw == adc units`? It should be ints. Might be worth saying it is adc units.

nulinspiratie:
Wait so raw is not the same as adc? So what's the difference?
And `raw2volts` also has type `float` as input, not `int`. Should this also be changed then?

And it's relevant for time tagging because we've decided to use volts there for the settings, as opposed to adc units

nulinspiratie:
Given how rare this case is, I'll close this PR. In the end it boils down to dividing by 4096. Feel free to reopen if you think it should be added

PR #249, opened 2024-11-21
Make qualang-tools compatible with python 3.12

TheoLaudatQM:
Make qualang-tools compatible with python 3.12

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B,B950
-ignore = E203, E501, W503, E741, B950, F405, F403
+ignore = E203, E501, W503, E741, B950, F405, F403, E231, E713, W604, E241, E221, E222, E225
 exclude = **/tests/**
 per-file-ignores =
     qualang_tools/config/components.py:E743
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B,B950
-ignore = E203, E501, W503, E741, B950, F405, F403
+ignore = E203, E501, W503, E741, B950, F405, F403, E231
 exclude = **/tests/**
 per-file-ignores =
     qualang_tools/config/components.py:E743
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 88
 select = C,E,F,W,B,B950
-ignore = E203, E501, W503, E741, B950, F405, F403, E231
+ignore = E203, E501, W503, E741, B950, F405, F403
 exclude = **/tests/**
 per-file-ignores =
     qualang_tools/config/components.py:E743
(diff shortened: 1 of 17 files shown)


TheoLaudatQM:
Hey @yomach, now that qm-qua support python 3.12, we need to release the constraint on qualang-tools as well.
Is there anything more that updating the pyproject.toml file and test everything?

yomach:
@TheoLaudatQM Yes, see how I solved it.
I also fixed/improved a few others, and did `poetry update`.
The only thing left is to test that everything works

yomach (approved):

TheoLaudatQM:
Nice, thanks a lot Yoav!
If think that the unit tests were performed with Python 3.10, is there a way to automate the tests to use both Python 3.8 and 3.12?

yomach:
Yes, but for that I would ask for @yonatanrqm help

nulinspiratie:
We test for multiple python versions in QUAM as well, I'll add it here too

nulinspiratie:
By the way, should we continue supporting 3.8? Python stopped support for 3.8

yomach:
> By the way, should we continue supporting 3.8? Python stopped support for 3.8

We will drop support for Python 3.8 soon (either qm-qua 1.2.2 or 1.2.3)

nulinspiratie:
There are a number of new linting errors (not real errors) due to pylint needing to be upgraded for Python 3.12. I ignore E231 but there are several more. Should we ignore all of them or resolve them? @yomach @TheoLaudatQM

nulinspiratie:
> I would naively say that we should resolve it since it is a requirement for python 3.12, even if it is just linting and without thinking too much I am guessing that there is a good reason behind adding these additional linting checks, but I also don't know more than that so I let you guys decide since you have more experience than me

I propose we do it in two stages. First we ignore the linting rules and verify that all tests are still working, then merge.
In a follow-up PR we fix all the linting issues. This way the changes in this PR don't explode, and the changes in the other PR are solely focused on linting issues

yomach:
I would just fix it, most of them are simple whitespace anyway.
Ignoring is a sure way to never fix it :)

nulinspiratie:
Ok that works for me as well. I will probably have time friday to handle this

nulinspiratie:
@yomach @TheoLaudatQM I can't believe it, it finally worked 🥲 
I fixed many linting errors. 
Additionally I had to use the specific Flake8 version 5.0.4 rather than the latest version 7.1.1 as this is the latest version that supports Python 3.8. As a result, I had to disable the following flake8 errors: `E231, E713, W604, E241, E221, E222, E225` because they are actually erroneous, and are removed in later Flake8 versions. Once we drop support for Python 3.8 we should remove these disabled errors again.

@TheoLaudatQM @yomach ready to merge?

yomach:
Nice work!!!
Yes, I'll merge.
I think that soon (next version?) we'll remove Python 3.8, and then we should revisit this lint part.

PR #251, opened 2024-11-30
Fix qcodes driver for qm-qua==1.2.1

TheoLaudatQM:
Fix the OPX qcodes driver to be compatible with qm-qua 1.2.1 (apparently the serialized stream processing format change a bit compared to previous versions)


TheoLaudatQM:
Yes agree...
I just added a simple check `if hasattr(gene.values[2].list_value.values[1], "list_value"):` I think that before 1.2.1 we always had a list_value, but maybe for simplification, now we can have directly a string_value if there is nothing else after...

PR #257, opened 2025-01-15
Fix/allow drag with alpha equals 0

TheoLaudatQM:
- Remove deprecated `delta` parameter
- Allow to generate a DRAG waveform even if alpha=0 per Tom's request

### DIFF
--- a/qualang_tools/config/waveform_tools.py
+++ b/qualang_tools/config/waveform_tools.py
@@ -40,6 +40,10 @@ def drag_gaussian_pulse_waveforms(
     if anharmonicity != detuning:
         # The complex DRAG envelope:
         z += 1j * gauss_der_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
+    elif alpha != 0:
+        raise ValueError(
+            "The complex envelop for the DRAG waveform cannot be created if anharmonicity = detuning and alpha != 0."
+        )
     # The complex detuned DRAG envelope:
     z *= np.exp(1j * 2 * np.pi * detuning * t * 1e-9)
     I_wf = z.real.tolist()  # The `I` component is the real part of the waveform
@@ -78,6 +82,10 @@ def drag_cosine_pulse_waveforms(amplitude, length, alpha, anharmonicity, detunin
     if anharmonicity != detuning:
         # The complex DRAG envelope:
         z += 1j * sin_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
+    elif alpha != 0:
+        raise ValueError(
+            "The complex envelop for the DRAG waveform cannot be created if anharmonicity = detuning and alpha != 0."
+        )
     # The complex detuned DRAG envelope:
     z *= np.exp(1j * 2 * np.pi * detuning * t * 1e-9)
     I_wf = z.real.tolist()  # The `I` component is the real part of the waveform
--- a/qualang_tools/config/waveform_tools.py
+++ b/qualang_tools/config/waveform_tools.py
@@ -26,46 +26,24 @@ def drag_gaussian_pulse_waveforms(
     :return: Returns a tuple of two lists. The first list is the 'I' waveform (real part) and the second is the
         'Q' waveform (imaginary part)
     """
-    delta = kwargs.get("delta", None)
-    if delta is not None:
-        print("'delta' has been replaced by 'anharmonicity' and will be deprecated in the future. ")
-        if alpha != 0 and delta == 0:
-            raise Exception("Cannot create a DRAG pulse with `anharmonicity=0`")
-        t = np.arange(length, step=1e9 / sampling_rate)  # An array of size pulse length in ns
-        center = (length - 1e9 / sampling_rate) / 2
-        gauss_wave = amplitude * np.exp(-((t - center) ** 2) / (2 * sigma**2))  # The gaussian function
-        gauss_der_wave = (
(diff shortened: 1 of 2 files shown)


TheoLaudatQM:
@KevinAVR I don't remember why this condition was added, do you know?

KevinAVR (approved):
@TheoLaudatQM I don't recall but the previous and later functions return the same result isn't it? when alpha = 0 then 1j*0

yomach (changes requested):
I think alpha=0 was supported before, so I'm not sure about this PR...
But in any case, if we remove a deprecated parameter, then:
1. Write it in the changelog
2. We need to increase the version!

TheoLaudatQM:
@yomach before you could only set  a detuning when alpha != 0 which led to confusions at the IQCC, now I just allowed it.
I updated the changelog including the deprecation change and yes we'll increase the version number when releasing it

qualang_tools/config/waveform_tools.py
```diff
@@ -26,46 +26,24 @@ def drag_gaussian_pulse_waveforms(
...
-        z = gauss_wave + 1j * 0
-        if alpha != 0:
-            # The complex DRAG envelope:
-            z += 1j * gauss_der_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
-            # The complex detuned DRAG envelope:
-            z *= np.exp(1j * 2 * np.pi * detuning * t * 1e-9)
-        I_wf = z.real.tolist()  # The `I` component is the real part of the waveform
-        Q_wf = z.imag.tolist()  # The `Q` component is the imaginary part of the waveform
+    if alpha != 0 and anharmonicity == 0:
+        raise Exception("Cannot create a DRAG pulse with `anharmonicity=0`")
```
yomach, line 30: @TheoLaudatQM But you see raise an error in this case, what am I'm missing?

qualang_tools/config/waveform_tools.py
```diff
@@ -26,46 +26,24 @@ def drag_gaussian_pulse_waveforms(
...
+        amplitude * (-2 * 1e9 * (t - center) / (2 * sigma**2)) * np.exp(-((t - center) ** 2) / (2 * sigma**2))
+    )  # The derivative of gaussian
+    if subtracted:
+        gauss_wave = gauss_wave - gauss_wave[-1]  # subtracted gaussian
+    z = gauss_wave + 1j * 0
+    if anharmonicity != detuning:
+        # The complex DRAG envelope:
+        z += 1j * gauss_der_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
+    # The complex detuned DRAG envelope:
+    z *= np.exp(1j * 2 * np.pi * detuning * t * 1e-9)
```
yomach, line 44: You moved this outside of the if statement, it's ok because if `detuning=0` then it doesn't matter?

qualang_tools/config/waveform_tools.py
```diff
@@ -88,42 +66,22 @@ def drag_cosine_pulse_waveforms(amplitude, length, alpha, anharmonicity, detunin
...
-        z = cos_wave + 1j * 0
-        if alpha != 0:
-            # The complex DRAG envelope:
-            z += 1j * sin_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
-            # The complex detuned DRAG envelope:
-            z *= np.exp(1j * 2 * np.pi * detuning * t * 1e-9)
-        I_wf = z.real.tolist()  # The `I` component is the real part of the waveform
-        Q_wf = z.imag.tolist()  # The `Q` component is the imaginary part of the waveform
+    if alpha != 0 and anharmonicity == 0:
+        raise Exception("Cannot create a DRAG pulse with `anharmonicity=0`")
```
yomach, line 70: Same

qualang_tools/config/waveform_tools.py
```diff
@@ -88,42 +66,22 @@ def drag_cosine_pulse_waveforms(amplitude, length, alpha, anharmonicity, detunin
...
+    end_point = length - 1e9 / sampling_rate
+    cos_wave = 0.5 * amplitude * (1 - np.cos(t * 2 * np.pi / end_point))  # The cosine function
+    sin_wave = (
+        0.5 * amplitude * (2 * np.pi / end_point * 1e9) * np.sin(t * 2 * np.pi / end_point)
+    )  # The derivative of cosine function
+    z = cos_wave + 1j * 0
+    if anharmonicity != detuning:
+        # The complex DRAG envelope:
+        z += 1j * sin_wave * (alpha / (2 * np.pi * anharmonicity - 2 * np.pi * detuning))
+    # The complex detuned DRAG envelope:
```
yomach, line 81: same

yomach (approved):

qualang_tools/config/waveform_tools.py
TheoLaudatQM: This is if anharmonicity is 0, not detuning

qualang_tools/config/waveform_tools.py
TheoLaudatQM: yes exactly and detuning can be 0 no matter the value of alpha

qualang_tools/config/waveform_tools.py
TheoLaudatQM: This is if anharmonicity is 0, not detuning

qualang_tools/config/waveform_tools.py
TheoLaudatQM: This can be derived if detuning != anharmonicity, but this is the only relevant condition, but I should raise an error otherwise

PR #263, opened 2025-02-06
Fix/2q rb

deanpoulos:
- Swap the order of the `circuit_depth` and `repeat` axes in both the QUA program and analysis, leading to more sensible results and reduced dependence on the circuit-depth density.
- Fixed the extraction of the fidelity by correcting the initial guess
- Improve the plotting of the decay curve by adding error bars at one-sigma
- Added feature to plot the state distribution as it evolves over circuit depth for debugging leakage.

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+
+## [0.18.3] - 2025-02-07
 ### Added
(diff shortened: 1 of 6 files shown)


yomach (commented):
It's a bit out of the scope, but we don't want examples under the "qualang_tools" folder, we have a separate example in the root of the git.
Can you move it there as part of this PR?

deanpoulos:
> It's a bit out of the scope, but we don't want examples under the "qualang_tools" folder, we have a separate example in the root of the git. Can you move it there as part of this PR?

Done

yomach (approved):
Ok from my side, might want Theo as well

deanpoulos:
> Ok from my side, might want Theo as well

Synced with Theo over call

PR #269, opened 2025-02-21
Feat/XArray data fetcher

nulinspiratie:
Adds a live data fetcher that automatically converts to xarray.

## Status
- [x] Working basic version
- [x] Live data acquisition
- [x] Support non-multiplexed readout (some qubits have not yet streamed any data out)
- [x] Concatenation of outer qubit sweep axis (`state1`, `state2`, .. -> `state`)
- [ ] Testing on multiple nodes
- [ ] Documentation

It's too early for a full review, but some early feedback on the user code would be good. Here is a screenshot of the code, including the different way in which sweep axes can be passed
Screenshot 2025-02-21 at 20 05 27


nulinspiratie:
Closing, moved to Qualibration-Libs repo

PR #273, opened 2025-03-04
Fix/octave tools

TheoLaudatQM:
Fix bug in Octave tools


MichalGQM (approved):
Thanks for fixing it!

PR #275, opened 2025-03-05
runs-on: ubuntu-22.04

TheoLaudatQM:
Update GitHub workflow with `runs-on: ubuntu-22.04`

### DIFF
--- a/.github/workflows/on-released.yml
+++ b/.github/workflows/on-released.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   publish:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/on-released.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   publish:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
yomach, line 9: I would go either to 24.04 or just to the latest.
Let's delay the next deprecation :)

TheoLaudatQM: Good point, for some reason I thought that 22.04 was the latest...
Done :)

yomach (approved):

PR #278, opened 2025-03-07
Change: DataHandler save key 'parents' -> 'ancestors'

nulinspiratie:
The datahandler adds a key called `parents` to `node.json`, which is a list containing one element (the previous idx). This was added so in the future we could allow for more complex timelines that include branches and merging.

However, it seems more appropriate to use the term `ancestors` for two reasons:
- In QUAlibrate, I want to reserve an additional keyword `parent` to indicate when a node is run as part of a graph. In this case, having both `parents` and `parent` can be confusing
- When talking about the timeline, it seems more natural to talk about ancestors and descendants

Technically speaking this is a breaking change, but I highly doubt anyone uses this so far

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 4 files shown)


nulinspiratie:
After further discussions, this will create too many breaking changes. Closing the PR.

PR #280, opened 2025-04-07
Feat: Add waveform report to data handler

nulinspiratie:
This PR adds support for the waveform report in the data handler. Usage:

```python
# We assume that `job` already exists as the output of a simulation
samples = job.get_simulated_samples()
wf_report = job.get_simulated_waveform_report()

results = {
    "waveform_report": wf_report,
    "samples": samples,  # Optional, allows the waveform report to be saved with specific simulation samples
}
```
This will save the waveform report both as a serialised dictionary, as well as an HTML file containing the interactive plot.


TheoLaudatQM (commented):
All good, just need to update the changelog :)

PR #285, opened 2025-05-12
Fix/multi_user

TheoLaudatQM:
Fixes the qm_session tool to work on both the OPX+ and OPX1000.
The logging filtering works well for the OPX+, but not so much for the OPX1000 as I can't get rid of "PHYSICAL CONFIG ERROR...", but it is a minor issue I think


paulQM (approved):
all good

PR #286, opened 2025-05-13
Features/config helpers

TheoLaudatQM:
Add helper functions for building the config/quam:

- Get the MW fem bands from the desired frequency
- Get the full scale power in dBm and wf amplitude from the desired power
- Get the Octave gain and wf amplitude from the desired power


paulQM (approved):
LGTM! Couldn't find nothing wrong with it

PR #288, opened 2025-05-15
Feat: Add support for plotly figure data saving

nulinspiratie:
This PR adds support for saving plotly figures using the data handler.
This is needed, among other things, by the video mode dash application


TheoLaudatQM (commented):
@nulinspiratie I just quickly went over it and it looks good, although I believe that it should come with a set of tests to ensure compatibility with future versions.
WHat do you think?

PR #289, opened 2025-05-16
Fix/voltage gate sequence

TheoLaudatQM:
Fix an inaccuracy due to the fixed point precision when deriving the compensation pulse for long fixed lengths.
I tested it and the fix was also confirmed by the customer.

### DIFF
--- a/qualang_tools/voltage_gates/voltage_gate_sequence.py
+++ b/qualang_tools/voltage_gates/voltage_gate_sequence.py
@@ -335,7 +335,8 @@ class VoltageGateSequence:
                     amplitude = declare(fixed)
                     eval_average_power = declare(int)
                     assign(eval_average_power, self.average_power[i])
-                    assign(amplitude, -Cast.mul_fixed_by_int(0.0009765625 / duration, eval_average_power))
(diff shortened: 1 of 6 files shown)


qualang_tools/voltage_gates/voltage_gate_sequence.py
```diff
@@ -335,7 +335,8 @@ def add_compensation_pulse(self, max_amplitude: float = 0.49, **kwargs) -> None:
                     amplitude = declare(fixed)
                     eval_average_power = declare(int)
                     assign(eval_average_power, self.average_power[i])
-                    assign(amplitude, -Cast.mul_fixed_by_int(0.0009765625 / duration, eval_average_power))
+                    assign(amplitude, -Cast.mul_fixed_by_int(0.01 / duration, eval_average_power))
+                    assign(amplitude, amplitude * 0.09765625)
```
yomach, line 339: This looks like a magic number, can you write the calculation for it?
Either directly or in a comment.

TheoLaudatQM: It is not really magic, it corresponds to `1/1024` to compensate for the multiplication of the average voltage by 1024 to avoid digitization issues as it must be an int.
I am adding it as a comment

PR #293, opened 2025-06-04
Feature/wirer html visualization

deanpoulos:
Implements:
- web-based
- scrollable
- compact/stacked
- saveable
- local

Visualization by default, to avoid having many figures when using multi-chassis.

image


TheoLaudatQM (approved):
I didn't dive into the code, but quickly tested it on the examples from qua-libs and it looks great!
Well done!

deanpoulos:
> I didn't dive into the code, but quickly tested it on the examples from qua-libs and it looks great! Well done!

Thanks!

PR #295, opened 2025-07-09
Fix/voltage gate sequence

TheoLaudatQM:
1. Remove the error due to the bug regarding the sampling rate
2. Play the compensation pulse using ramps to avoid overflow issues
3. Add parametrization of the base pulse amplitude based on the output mode (amplified: 0.5V; else: 0.25V) and the corresponding power of 2 for bitshifting
4. Add semi-automatic testing


qualang_tools/voltage_gates/voltage_gate_sequence.py
```diff
@@ -295,18 +296,17 @@ def add_compensation_pulse(self, max_amplitude: float = 0.49, **kwargs) -> None:
...
-                    operation = self._add_op_to_config(gate, "compensation", amplitude=0.25, length=16)
                     eval_average_power = declare(int)
                     comp_duration = declare(int)
                     duration_4ns = declare(int)
                     duration_4ns_pow2 = declare(int)
                     duration_4ns_pow2_cur = declare(int)
                     amplitude = declare(fixed)
                     # Exact duration of the compensation pulse
-                    # take into account a gap of 96ns for the derivation of the compensation pulse
+                    # take into account a gap of 110ns for the derivation of the compensation pulse
```
yomach, line 306: It doesn't depend on the version? it's just constant?

PR #302, opened 2025-09-11
Fix: callable_from_qua support for qm-qua 1.2.3

nulinspiratie:


nulinspiratie:
@TheoLaudatQM since qm-qua 1.2.3 is now required in py-qua-tools, it no longer requires a breaking change.
Also, I believe it should be compatible with both opx1000 and opx+, afaik there isn't anything in here that's specific to opx1000.
Also fixed the changelog message.

TheoLaudatQM (approved):

PR #306, opened 2025-09-26
Wirer update for digital channels

TheoLaudatQM:
Add the possibility to attach a digital marker to a DC channel and to set up digital output only channels for triggering lasers for instance.
I also added the corresponding tests.

### DIFF
--- a/qualang_tools/wirer/wirer/wirer.py
+++ b/qualang_tools/wirer/wirer/wirer.py
@@ -119,7 +119,12 @@ def allocate_dc_channels(spec: WiringSpec, instruments: Instruments):
     """
     Try to allocate DC channels to an LF-FEM or OPX+ to satisfy the spec.
     """
-    dc_specs = [ChannelSpecLfFemSingle(), ChannelSpecOpxPlusSingle()]
+    dc_specs = [
+        # LF-FEM, Single analog output
+        ChannelSpecLfFemSingle() & ChannelSpecLfFemDigital(),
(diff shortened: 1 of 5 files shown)


qualang_tools/wirer/wirer/wirer.py
```diff
@@ -119,7 +122,12 @@ def allocate_dc_channels(spec: WiringSpec, instruments: Instruments):
     """
     Try to allocate DC channels to an LF-FEM or OPX+ to satisfy the spec.
     """
-    dc_specs = [ChannelSpecLfFemSingle(), ChannelSpecOpxPlusSingle()]
+    dc_specs = [
+        # LF-FEM, Single analog output
+        ChannelSpecLfFemSingle() & ChannelSpecLfFemDigital(),
+        # OPX+, Single analog output
+        ChannelSpecOpxPlusSingle() & ChannelSpecOpxPlusDigital(),
+    ]
```
NicolasStaudenmaier, line 131: Is there a reason to have only LF-FEM and not MW-FEM?

NicolasStaudenmaier:
I tested to add a digital only channel. Filtering by the wiring specs at channel allocation filters out the channel template if it is not triggered. Corresponding line is l.32 in `wirer/connectivity/channel_spec`: 
`digital_channel_only_if_triggered = channel.signal_type == "analog" or wiring_spec.triggered`

deanpoulos (approved):
Looks great! As long as tests pass I'm happy

PR #313, opened 2025-10-31
Replace fetch_all with fetch_results

4hsiangQM:
To implement the new `fetch_results` function in `qm-qua` > 1.2.3. I replace the `fetch_all` function.

After this change, I have run all the node in `~/Quantum-Control-Applications/Superconducting/Two-Fixed-Coupled-Transmons` All programs test on opx+ and opx1000.

### DIFF
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -28,13 +28,11 @@ class fetching_tool:
         self.data_list = data_list
         self.mode = mode
         self.results = []
-        self.data_handles = []
         self.res_handles = job.result_handles
         if mode == "live":
             for data in self.data_list:
                 if hasattr(self.res_handles, data):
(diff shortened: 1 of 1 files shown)


yomach (commented):
Hi @4hsiangQM - Thanks!
It looks good, few questions/comments:
1. Did you test it? Both on OPX+ and OPX1000?
2. I see self.data_handles is no longer being used besides lines 36-37, so we can replace them with: `self.res_handles.get(data).wait_for_values(1)` and remove self.data_handles.

4hsiangQM:
hi @yomach,
1. I’ve tested all the files in the` Two-Fixed-Coupled-Transmons` folder on both the OPX+ and OPX1000, and everything looks good to me.
2. it is good point. I will make a change. thanks for the comment.

PR #317, opened 2025-11-21
Fix/ ramp all elements in sequence to 0V before calculating compensation pulse

ChloeBO-QM:
Changed the behavior of `add_compensation_pulse()` so that the voltage on all elements of the sequence are ramped to 0V before the compensation pulse is calculated. This way, if the calculation of the compensation pulse introduces gaps, it will have little effect on the integrated voltage of the pulse and the compensation will be more accurate. Also, the voltages are all ramped to 0V after the compensation pulse has been applied to avoid any gaps after the compensation pulse (e.g. before `ramp_to_zero`) from affecting the integrated voltage of the sequence.

- CHANGELOG has been updated.
- README has been updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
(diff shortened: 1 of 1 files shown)


OziEgri (approved):
LGTM

PR #318, opened 2025-12-01
Modify manual output control to make it compatible with OPX1000

cjwu-qm:
Update the manual output control python file so that it is compatible with OPX1000.
Also updated the manual so that it includes the new compatibility.


OziEgri (changes requested):

cjwu-qm:
## Summary
- Remove stale commented-out QM cleanup code in `ManualOutputControl.__init__` to keep the shutdown path clear and maintainable.
- Replace constructor `print(...)` with module logging (`logger.info(...)`) so runtime messaging can be managed by application logging configuration.
- Extract repeated OPX/OPX+ port-index arithmetic into helper functions (`compute_con_number`, `compute_port_in_controller`) and reuse them across analog/digital config generation.
- Introduce `DEFAULT_INTERMEDIATE_FREQUENCY_HZ` and replace hard-coded intermediate-frequency literals (`0e6` / `0.0`) with the named constant for consistency and readability.
- Preserve existing behavior and config structure while cleaning up implementation details (including replacing identity comparison with value comparison in controller checks).

cjwu-qm:
@OziEgri I have modified according to your comments. Please let me know if you have further comments! Thank you.

PR #319, opened 2025-12-07
Fetching tool add backward compatibility

4hsiangQM:
I've added backward compatibility to choose which fetching method is being used by `_query_multiple_streams_fetching_capability` function. Please review it when you have time. thanks!

### DIFF
--- a/qualang_tools/results/results.py
+++ b/qualang_tools/results/results.py
@@ -26,7 +26,7 @@ class fetching_tool:
             raise Exception("The provided data list is empty.")
         if mode not in ["live", "wait_for_all"]:
             raise Exception(f"Mode '{mode}' is not supported. Supported modes are ['live', 'wait_for_all']")
-        
+
         self.data_list = data_list
         self.mode = mode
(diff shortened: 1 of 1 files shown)


qualang_tools/results/results.py
```diff
@@ -80,14 +90,48 @@ def fetch_all(self):
 
         :return: all result of current result stream as a list of python variables
         """
-        if self.mode == "wait_for_all":
-            results = self.res_handles.fetch_results(wait_until_done=True, stream_names=self.data_list)
-            self.results = [results.get(data) for data in self.data_list]
-        elif self.mode == "live":
-            results = self.res_handles.fetch_results(wait_until_done=False, stream_names=self.data_list)
-            self.results = [results.get(data) for data in self.data_list]
+        if self.multiple_streams_fetching_cap:
```
OziEgri: @4hsiangQM What is the difference between "Wait_for_all" and self.multiple_streams_cap ? why do we need to check the cap first?

PR #336, opened 2026-03-26
Feature/sweep post process

natangqm:
## Summary

Adds `fetch_xarray_data` — fetches results from a completed QUA job and returns them as an `xarray.Dataset`, with each result stream automatically shaped and labelled according to the QUA iterables (sweep) structure. Native iterables are reassembled from their per-index streams, and axes are transposed so the Dataset's dimension ordering matches the original iteration order.

## qm-qua compatibility

The feature relies on qm-qua APIs introduced in **1.3.1** (e.g. `STREAM_NAME_SEPARATOR`, `declare_with_stream`, the `qua_iterators` extensions). Rather than raise the package-wide floor and break users on older qm-qua, the requirement is handled gracefully:

- The qm-qua imports are **deferred** out of module scope (type hints under `TYPE_CHECKING` + `from __future__ import annotations`), so `import qualang_tools` works on any supported qm-qua.
- `fetch_xarray_data` is **version-gated**: on qm-qua < 1.3.1 it raises a clear `ImportError("fetch_xarray_data requires qm-qua>=1.3.1 ...")` instead of an opaque import crash.
- `MIN_QM_QUA_VERSION = "1.3.1"` is the single source of truth, driving the runtime gate, error message, and test skips.

The `qm-qua` floor in `pyproject.toml` stays at `>=1.2.3`. Because qm-qua 1.3.1 dropped Python 3.9, the lock resolves per-interpreter: **1.3.1 on Python 3.10+, 1.2.6 on Python 3.9** — 3.9 support is preserved.

## Changes

- **Add** `qualang_tools.results.fetch_xarray_data` (+ README in `qua_iterables_processing/`).
- **Deps:** `xarray` promoted from optional extra to a core dependency; `packaging` added as an explicit core dependency (used for the version comparison).
- **Tests:** full coverage of basic/averaging/zip/edge-case behaviour, plus version-guard tests. Feature tests are excluded from collection on qm-qua < 1.3.1 via `collect_ignore` (their top-level qm-qua 1.3.1 imports can't be collected on older qm-qua, so a `pytest.mark.skipif` is insufficient).
- **CHANGELOG.md** updated.

## CI

Green across the matrix: Python 3.10/3.12/3.13 install qm-qua 1.3.1 and **run** the feature tests; Python 3.9 installs 1.2.6 and cleanly skips them.


qualang_tools/loops/qua_iterable_postprocess.py
```diff
@@ -0,0 +1,141 @@
...
+
+from qm.api.v2.job_api.job_api import JobApi
+from qm.qua.extensions.qua_iterators.qua_iterators_base import IterableBase
+from qm.qua.extensions.qua_iterators.qua_native_iterators import NativeIterableBase
+from qm.qua.extensions.qua_iterators import QuaProduct
+from qm.qua.extensions.qua_iterators.qua_zip import QuaZip, NativeZipIterable
+from qm.qua._dsl.stream_processing.direct_stream_processing import STREAM_NAME_SEPARATOR
+
+
+def _is_native(itr: IterableBase) -> bool:
```
yonatanrqm: This logic is sensitive to changes in the implementation of the iterables.
Not so much to do about it, unless you have an idea.

tests/tests_qua_utilities/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+        action="store",
+        default="v3_6_0",
+        help="Cloud simulator QOP version to use for the tests - e.g., v3_5_0, v3_6_0, default: local for local simulator"
+    )
+    parser.addoption(
+        "--cloudsim-pwd",
+        action="store",
+        default="jG4yUA9YTqcYHBVsVenf",
+        help="Cloud simulator password"
+    )
```
nulinspiratie, line 96: does it make sense to add the pwd? and email? Seems risky

qualang_tools/results/sweep_program_processing/qua_iterable_postprocess.py
```diff
@@ -0,0 +1,150 @@
...
+        current_order = native_itr + [itr for itr in non_avg_itr if itr.is_qua_iterable]
+        perm = [current_order.index(itr) for itr in non_avg_itr]
+        stream_data[stream_name] = np.transpose(result, perm)
+
+    return stream_data
+
+
+def fetch_xarray_data(
+    job: JobApi,
+    sweep: Union[QuaProduct, Sequence[IterableBase]],
```
nulinspiratie, line 87: One thing I'm wondering is whether the sweep needs to be added here. Would it be possible to instead include it as part of the program or job? The reason I'm asking is that one might run into a situation where the user doesn't have easy access to the sweeps, and thus won't be able to retrieve the xarray dataset

qualang_tools/results/sweep_program_processing/qua_iterable_postprocess.py
```diff
@@ -0,0 +1,150 @@
+from typing import Optional, Sequence, Union
```
nulinspiratie, line 1: Should also add relevant functions to `__init__.py`

qualang_tools/results/sweep_program_processing/qua_iterable_postprocess.py
nulinspiratie: I should say that this isn't a hard requirement, and if it's possible it's something for a follow-up release

tests/tests_qua_utilities/test_fetch_xarray_zip.py
```diff
@@ -0,0 +1,53 @@
...
+        QuaIterable("frequency", frequencies),
+    ])
+    with program() as prog:
+        for args in prod:
+            s = declare_with_stream(int, "val_st")
+            assign(s, args.shot)
+
+    xarray_data = simulate_and_fetch(qmm, prog, prod)
+
+    assert_dims_and_shape(xarray_data, "val_st", ("shot", "qb_amp", "frequency"), (10, len(native_zip_values), len(frequencies)))
```
nulinspiratie, line 28: Note that xarray supports having multiple axes for a single dimension. I'm not sure whether that's implemented here, but I think it's the right approach. So whenever multiple axes are zipped together, there will be a single dimension with multiple Xarray coordinate axes attached to it

nulinspiratie (changes requested):
I think this functionality better belongs in `results` folder because it is similar to fetching tool, which is located there.
Additionally, there is the [XarrayDataFetcher](https://github.com/qua-platform/qualibration-libs/tree/main/qualibration_libs/data) in `Qualibration-Libs`. I think the behavior of this tool should mimic that one, i.e. be used as a context manager that can also retrieve data while still acquiring. In fact, we should consider merging the functionality of this feature and the XarrayDataFetcher into one tool.

It's fine if the current version doesn't support data fetching while acquiring, but I think we should at least match the UX.

Since XarrayDataFetcher is in QualibrationLibs, we should think together with @OziEgri where the final tool should reside, py-qua-tools or qualibration-libs? I think it makes sense to have one repository for this and related tools

The PR is also missing a README, I'm sending you a draft separately @natangqm

nulinspiratie:
[README.md](https://github.com/user-attachments/files/26741573/README.md)
Here's the proposed readme to accompany it @natangqm

tests/tests_qua_utilities/conftest.py
natangqm: Its like this in m-qua (both in conftest and as secret), but I forgot that pyquatools is open source. will move it to be secret

qualang_tools/results/sweep_program_processing/qua_iterable_postprocess.py
natangqm: Its not part of the design. I think that adding it to the job might be too big of a jump. Lets see if after clients using this feature they will have any issues with the design

tests/tests_qua_utilities/test_fetch_xarray_zip.py
natangqm: This what should happen here. If this is not what's happening let me know

nulinspiratie:
I also propose we ditch the naming "sweep program", it could confuse customers. I propose to instead move the contents to the file `qualang_tools.results.xarray_results.py`, wdyt @natangqm ?

nulinspiratie (approved):
Approved, but minor docs comments to implement before merging

qualang_tools/results/qua_iterables_processing/qua_iterable_postprocess.py
```diff
@@ -0,0 +1,150 @@
...
+    original iteration order. Each data variable in the returned Dataset is indexed only
+    by the dimensions that were not stream-averaged for that variable. Coordinate metadata
+    (e.g. units) from the iterables is attached to the corresponding Dataset coordinates.
+
+    Args:
+        job: A completed QUA job from which to fetch results.
+        iterables: The QUA iterables used in the program — either a ``QuaProduct``
+            or a sequence of ``IterableBase`` objects.
+        wait_until_done: If True, block until the job completes before fetching results.
+            Defaults to False.
```
nulinspiratie, line 141: It is unclear here what wait_until_done=False actually does. Earlier in the docstring it already says that it retrieves result streams from a completed QUA job, so what happens if it's not complete? Does it acquire whatever is available, or something else?
This could be clarified in the docstring, and also the README.

qualang_tools/results/qua_iterables_processing/qua_iterable_postprocess.py
```diff
@@ -0,0 +1,150 @@
...
+        # Transpose so each dim sits at its original iteration position (among non-averaged only)
+        # Current order after reshape: all native first, then non-native non-averaged
+        current_order = native_itr + [itr for itr in non_avg_itr if itr.is_qua_iterable]
+        perm = [current_order.index(itr) for itr in non_avg_itr]
+        stream_data[stream_name] = np.transpose(result, perm)
+
+    return stream_data
+
+
+def fetch_xarray_data(
```
nulinspiratie, line 122: One thing to consider here is that from a user's perspective they would likely want to keep acquiring until the job is done, then do a final acquisition, and then use that as the final dataset, so something like:
```python
while not job.is_done():  # forgot the exact command here
    ds = fetch_xarray_data(job, iterables)
ds = fetch_xarray_data(job, iterables)  # Acquire one last time
```
This is a bit clunky, so we should come up with a better solution. For example, in the previous xarray data fetcher, it was done through a for loop, something like:
```python
for ds in fetch_data(job, iterables):  # again, forgot the exact command
    ...
# No need for a final fetch, the `for` loop logic already encapsulates this
```
I'm not saying this is the right way, but we should have something simple that avoids us having to implement the logic in the first example. We don't need to do it here, but we need a solution in the next phase

natangqm: we can do that in next version

qualang_tools/results/qua_iterables_processing/qua_iterable_postprocess.py
natangqm: Its clarified in the readme and in the docstring

yomach:
@OziEgri Any objections we merge this?

qualang_tools/results/qua_iterables_processing/qua_iterable_postprocess.py
```diff
@@ -1,12 +1,48 @@
...
+def _installed_qm_qua_version() -> str:
+    """Return the installed qm-qua distribution version."""
+    from importlib.metadata import version
+
+    return version("qm-qua")
+
+
+def qua_iterables_supported() -> bool:
+    """Whether the installed qm-qua is recent enough for QUA-iterable post-processing."""
+    from packaging.version import Version
```
yonatanrqm: Move upwards, and use the `Version` in the constant, and in `_installed_qua_version`

yonatanrqm (approved):

natangqm:
@OziEgri please review

OziEgri (commented):
LGTM, would be good to add Theo or Kevin to the review, as they have more experience here

PR #343, opened 2026-05-06
Bump python range to >=3.10,<3.15

thaihientran:
## Summary
- Bump Python range from `>=3.9, <=3.13` to `>=3.10, <3.15`. Drops 3.9; allows 3.13/3.14.
- Add a python marker (`>=3.10,<3.13`) on `qm-qua` because the latest public-PyPI `qm-qua` wheel (1.2.6) caps at `<3.13`. Until a `<3.15` `qm-qua` wheel is published to PyPI, `qm-qua` will not install on Python 3.13+. Marker should be dropped at that point.
- Tighten the lower-bound `scipy` marker from `>=3.9,<3.13` to `>=3.10,<3.13` to match.

## Test plan
- [ ] CI green
- [ ] Smoke import on Python 3.10–3.12 (3.13+ requires updated qm-qua publish)


yomach:
Done as part of https://github.com/qua-platform/py-qua-tools/pull/353

PR #346, opened 2026-07-23
Bump version to 0.22.1.dev1 and update changelog

ohad-lev-qm:


TheoLaudatQM (approved):
Approving for now, but I am under the impression that adding a new dev version for each PR is quite tedious and makes the changelog a bit messy.
Why not just mark the main as dev to indicate that the package hasn't been officially release and work freely on it?

PR #349, opened 2026-07-29
Add closest IF Octave calibration helpers

NicolasStaudenmaier:
## Summary

Adds support for applying the closest available Octave IF calibration from `calibration_db.json` when no exact IF match exists. LO and gain still need to be matched, then the calibration with the closest IF value can be applied. 

New functions in `octave_tools`:
- `get_closest_calibration_parameters_from_db` — lookup nearest IF calibration
- `set_closest_correction_parameters_to_opx` — apply it to a single element
- `apply_closest_calibrations` — apply it to all Octave elements in a config

Also refactors `get_calibration_parameters_from_db` to resolve the Octave port from either `RF_inputs` or `RF_outputs`.

## Test plan

- [x] Added unit tests in `tests/test_octave_tools.py`
- [x] Verified on hardware with a config where the target IF has no exact calibration entry. Checked the analog output on the SA.


OziEgri:
@NicolasStaudenmaier please take a look at the CI/Cd results above , they fail for what seems to be a format issue (did you run pylint and black before submitting?) thanks.

NicolasStaudenmaier:
@OziEgri now it's passing all checks. 
Also updated the changelog again because after the last merges the changes here appeared under the latest released version.

OziEgri (approved):
LGTM, thanks!

PR #350, opened 2026-08-12
Feature/add detuned xy drive

soonqm:
- fix bakery not working for amplitude list
- added WiringLineType for zz and xy_detuned
- make connectivity qubit-pair string consistent with [_parse_qubit_pair_ids](https://github.com/qua-platform/quam-builder/blob/71fffe7f309df61eb184560f752f7d29db80aae3/quam_builder/builder/quantum_dots/build_utils.py#L253)


qualang_tools/wirer/connectivity/wiring_spec.py
```diff
@@ -37,7 +35,8 @@ class WiringLineType(Enum):
     CHARGE = "q"
     COUPLER = "c"
     CROSS_RESONANCE = "cr"
-    ZZ_DRIVE = "zz"
+    ZZ = "zz"
```
Deepakkhurrana, line 38: **Blocking:** renaming `ZZ_DRIVE` → `ZZ` is a breaking API change.

Downstream (e.g. quam-builder `build_quam.py`) still uses `WiringLineType.ZZ_DRIVE`. Even though the value stays `"zz"`, the enum member name must either:
- keep a compatibility alias: `ZZ_DRIVE = ZZ`, or
- land a coordinated quam-builder PR in the same release.

qualang_tools/wirer/connectivity/element.py
```diff
@@ -36,7 +36,7 @@ class QubitPairReference:
     target_index: Union[int, str]
 
     def __str__(self):
-        return f"q{self.control_index}-{self.target_index}"
+        return f"q{self.control_index}-q{self.target_index}"
```
Deepakkhurrana, line 39: **Blocking / cross-repo:** changing `q{control}-{target}` → `q{control}-q{target}` is desirable for consistency with QD `_parse_qubit_pair_ids`, but superconducting builder currently does:
```python
qc, qt = qubit_pair_id.split("-")
qt = f"q{qt}"
```
So `q1-q2` becomes `control=q1`, `target=qq2.`

The  quantum-dots `_parse_qubit_pair_ids` helper gets this right  butSC builder should stay self-sufficient. Before (or with) merging this string change, quam-builder needs an SC-local pair-id parser (accepting both q1-2 and q1-q2) wired into superconducting build_quam / related pair parsing. Thoughts?

Until that lands, this change will break SC wiring → QUAM construction.

qualang_tools/bakery/bakery.py
```diff
@@ -1054,8 +1054,9 @@ def run(
                         index2 = list(zip(*amp_array))[0].index(qe)
                         amp = list(zip(*amp_array))[1][index2]
                         if type(amp) is list:
-                            raise TypeError("Amplitude can only be a number (either Python or QUA variable)")
-                        qua.play(f"baked_Op_{self._ctr}" * qua.amp(amp), qe)
+                            qua.play(f"baked_Op_{self._ctr}" * qua.amp(*amp), qe)
```
Deepakkhurrana, line 1057: Good fix, the old guard blocked the amplitude-matrix case this PR needs. Two things:

1. `type(amp) is list` misses tuples and numpy arrays, which fall through to the `else` and fail deeper in `qua.amp`. Consider using  `isinstance(amp, (list, tuple, np.ndarray))`.
2. No length check. `qua.amp` takes 1 or 4 args; a 2- or 3-element list now raises inside qm-qua instead of giving a clear error. Add `if len(amp) not in (1, 4): raise ValueError(...)`.

qualang_tools/wirer/connectivity/wiring_spec.py
```diff
@@ -37,7 +35,8 @@ class WiringLineType(Enum):
...
+    DETUNED_DRIVE = "xy_detuned"
```
Deepakkhurrana, line 39: **Nit / convention:** `DETUNED_DRIVE = "xy_detuned"` is inconsistent with the existing short-code style for SC line types (`"xy"`, `"rr"`, `"cr"`, `"zz"`, `"z"`, …). It’s the only long, underscored value in that set.

I see the upside, wiring paths use `line_type.value`, so `"xy_detuned"` lines up with the QUAM attribute name,  but if we want to keep the enum-value convention tight, prefer a short code (e.g. `"xd"`) and map it to `xy_detuned` in the builder.

Not a blocker if you intentionally chose path ↔ attribute alignment over abbreviation consistency; worth a brief note in the PR either way.

tests/wirer/test_wirer_cross_resonance.py
```diff
@@ -33,14 +36,16 @@ def test_2q_allocation_cross_resonance(instruments_2lf_2mw):
 
     for i, qubit_pair in enumerate(qubit_pairs):
         xy_channels = connectivity.elements[QubitReference(qubit_pair[0])].channels[WiringLineType.DRIVE]
+        xy_detuned_channels = connectivity.elements[QubitReference(qubit_pair[0])].channels[WiringLineType.DETUNED_DRIVE]
         cr_channels = connectivity.elements[QubitPairReference(*qubit_pair)].channels[WiringLineType.CROSS_RESONANCE]
         zz_channels = connectivity.elements[QubitPairReference(*qubit_pair)].channels[WiringLineType.ZZ_DRIVE]
```
Deepakkhurrana, line 41: A few test issues:

1. Still uses `WiringLineType.ZZ_DRIVE` after the rename to `ZZ`: this will `AttributeError`.
2. In the Octave test, `xy_detuned_channels` is included in the equality loop but never checked with `assert len(xy_detuned_channels) == ...` (unlike the MW-FEM test).
3. CI currently reports 0 tests for the head commit: can we confirm these wirer tests are actually collected/running?

Deepakkhurrana (changes requested):
Thanks @soonqm for fixes! Some changes before we merge this:


Commits mention CHANGELOG updates, but `[Unreleased]` doesn’t document:
- `DETUNED_DRIVE` / `add_qubit_detuned_drive_lines`
- `ZZ_DRIVE` → `ZZ` (breaking)
- `QubitPairReference` string format change (breaking)
- bakery amp-list fix

Please add those under Added/Changed/Fixed as appropriate.

qualang_tools/wirer/connectivity/wiring_spec.py
soonqm: ZZ is currently unreleased and no active user. Changing will not have impact

qualang_tools/wirer/connectivity/element.py
soonqm: Do we copy the same function as well? Or, I can revert this

qualang_tools/bakery/bakery.py
soonqm: 76abdb85271a38ceb674a7a9193b7e1a2c054b7f

qualang_tools/wirer/connectivity/wiring_spec.py
soonqm: 03adbf418a409670343e3d50728b45c6f6262552

tests/wirer/test_wirer_cross_resonance.py
soonqm: zz is not tested properly on any real system, any bug will need to defer until anyone is interested in using

zz_drive -> zz update
0fafc22995245f3eec0b4a52c7ccf8582e2c2c3b

qualang_tools/wirer/connectivity/element.py
Deepakkhurrana: I would say keep q{control}-q{target}, don’t revert this PR.

The SC quam builder still does qt = f"q{qt}", so q1-q2 becomes target qq2. That’s in quam-builder #146, build_quam.py (~118–119), not here.

Please fix it there the same way as QD _parse_qubit_pair_ids / modify_quam.py: split on -, add q only if it’s missing, on both qc and qt. Then old q1-2 and new q1-q2 both work.

PR #352, opened 2026-09-01
Remove deprecated 'version' key from QUA configs

yomach:
## Summary
- `qm-qua` warns whenever a QUA config dict contains a `"version"` key (deprecated since `1.2.2`, to be removed in `2.0.0`). `ConfigBuilder` (`QMConfiguration`), `ManualOutputControl`, and the config-GUI upload template all set this key unconditionally, so any user building a config through those paths hits the warning — not just the CI test that originally surfaced it.
- Removes the key from those library code paths, and from every test/example fixture that hardcoded it, so the warning can't resurface through a stale fixture.
- Drops the now fully unused `version` parameter from `QMConfiguration.__init__` — it had no callers anywhere in the repo and the old code already ignored whatever was passed, hardcoding `1` regardless.
- Adds a `CHANGELOG.md` entry under `[Unreleased]`.

## Test plan
- [x] `pytest tests/` (excluding `test_config_gui.py`, which needs the optional `dash`/`configbuilder` extra, and two pre-existing failures unrelated to this change: `test_config_builder.py::test_deprecated_warning` uses the removed `pytest.warns(None)` API, and `test_digital_filters.py` fails the same way before this change) — 463 passed, 0 warnings mentioning `'version'` or the Qua-config deprecation message.
- [x] Repo-wide `grep -rn '"version"'` confirms only the unrelated `config_structure["info"]["version"]` GUI-schema field remains.

Fixes the deprecation warning flagged by the `#sdk-deprecation-warnings` support-matrix bot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
   - `set_closest_correction_parameters_to_opx` applies those parameters to a single element.
   - `apply_closest_calibrations` applies the closest calibration to all Octave elements in a config.
 
+### Fixed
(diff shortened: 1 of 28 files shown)


OziEgri (approved):
LGTM, feel free to merge

PR #354, opened 2026-09-09
fix resultAnalysis name

ChloeBO-QM:
`prog.qua_program.result_analysis` has been changed to `prog.qua_program.resultAnalysis`. Updated `_get_stream_processing` method accordingly.


TorbjornRasmussenQM (approved):
Thanks for updating this, can you also add it to the changelog?

ChloeBO-QM:
> Thanks for updating this, can you also add it to the changelog?

Done.

